PR #22, opened 2019-11-15
ci: Run live tests in circleci

KitBurgess:


.circleci/config.yml
```diff
@@ -29,11 +29,17 @@ jobs:
             - ~/.cache/pip
 
       - run:
-          name: "Run tests"
+          name: "Run local tests only "
```
kujon: Local here refers to the mocked api?

.circleci/config.yml
```diff
@@ -29,11 +29,17 @@ jobs:
...
           command: |
             . .venv/bin/activate
             export SKIP_TAGS=real; nose2
 
+      - run:
+          name: "Run all tests"
+          command: |
+            . .venv/bin/activate
+            nose2
```
kujon: Does this mean that the CI is running the local tests twice?

.circleci/config.yml
KitBurgess: Yes

.circleci/config.yml
KitBurgess: Yes it does

.circleci/config.yml
KitBurgess:
```suggestion
          name: "Run local tests against mocked api"
```

.circleci/config.yml
kujon: Shall we not do that and have shorter build times?

KitBurgess: We could do, the local tests currently take <1second to run, so i don't think it's a problem right now.

KitBurgess:
Fat finger merged that prematurely by accident

Issue #29, opened 2019-11-18
Tried to pull out a reference df of all vessels but only got 200 results?


syed1992:
![image](https://user-images.githubusercontent.com/45287337/69049113-085b6180-09f7-11ea-8c5f-b5b4be308aa5.png)

PR #41, opened 2019-11-18
fix: Return all vessels from search, not just arbitrary 100

KitBurgess:


vortexasdk/client.py
```diff
@@ -41,32 +41,37 @@ def search(self, resource: str, **data) -> List:
...
 
-        return responses
+        send_request = lambda x: self._sent_post_request(url, payload, size, x)['data']
+
+        responses = list(pmap(send_request, offsets))
+
+        flattened = [x for y in responses for x in y]
+
+        assert len(flattened) == total
+        return flattened
```
KitBurgess: This is the key change here

KitBurgess:
This closes #29

PR #44, opened 2019-11-18
perf: Call API in parallel

KitBurgess:


tests/endpoints/test_cargo_movements_real.py
```diff
@@ -79,3 +80,16 @@ def test_search_list_filter_id(self):
...
+
+    def test_speed(self):
+        with Timer("Search") as t:
+            cms = CargoMovements().search(
+                filter_time_min="2019-08-29T00:00:00.000Z",
+                filter_time_max="2019-08-30T00:00:00.000Z",
+            )
+
+        with Timer("to_list") as t:
+            list = cms.to_list()
```
cvonsteg: `list` is a built-in function name, and should not be overwritten.  This could be called `list_of_cms`

vortexasdk/client.py
```diff
@@ -33,56 +24,68 @@ def get_reference(self, resource: str, id: ID) -> str:
...
-        responses = list(pmap(send_request, offsets))
+            responses = pool.map(functools.partial(_send_post_request_data, url=url, payload=payload, size=size),
+                                 offsets)
 
-        flattened = [x for y in responses for x in y]
+            flattened = [x for y in responses for x in y]
 
-        assert len(flattened) == total
-        return flattened
+            assert len(flattened) == total, f'len(flattened) = {len(flattened)}, total = {total}'
```
cvonsteg: Is a failure of this assertion meant to give the end-user any insight into the error, or is it just for developer debugging?  If it's the latter it's fine, but if it's intended for the end-user, I don't think message doesn't help much as it references variables created in the scope of the function.  In which case something like 'number of responses is does not match number of requests` may be more insightful.

PR #51, opened 2019-11-19
ci: Add git change log

KitBurgess:


scripts/export_package.sh
```diff
@@ -3,7 +3,8 @@ set -e
 
 . venv/bin/activate
 
-gitchangelog
+sudo gem install github_changelog_generator
```
kujon: Does ruby come preinstalled in your circle docker image?

KitBurgess: Nope, this export_package.sh script is semi-automated ;), only currently run in master branch by admins

PR #52, opened 2019-11-20
fix: Move to pytest, run mock setup before each test

KitBurgess:


KitBurgess:
Move to pytest because it's a far bigger project than nose, and has more descriptive visuals 
image

PR #63, opened 2019-11-22
feat: Add vessel movements endpoint

KitBurgess:


vortexasdk/endpoints/vessel_movements.py
```diff
@@ -0,0 +1,113 @@
...
+        |  0 | 2017-09-30T15:30:27+0000 | 2017-10-03T01:46:06+0000 |  9.21091e+06 | ADEBOMI 3     | handysize             | Netherlands                     | Rotterdam [NL]               | Netherlands                          | Rotterdam [NL]                    |                  nan | nan                             |
+        |  1 | 2017-08-29T14:51:32+0000 | 2017-10-04T14:46:21+0000 |  9.64544e+06 | AEGEAN VISION | suezmax               | Netherlands                     | Rotterdam [NL]               | Singapore                            | Singapore [SG]                    |               852261 | High Sulphur                    |
+
+        [Vessel Movements Endpoint Further Documentation](https://docs.vortexa.com/reference/POST/vessel-movements/search)
+
+        """
+        geog = lambda x: convert_to_geography_ids(to_list(x))
+        corporation = lambda x: convert_to_corporation_ids(to_list(x))
+        ves = lambda x: convert_to_vessel_ids(to_list(x))
+        product = lambda x: convert_to_product_ids(to_list(x))
```
cvonsteg: `nit:` named lambdas are considered un-pythonic.  In this case I'd suggest just creating the variables you want to pass to the `params` dict here, directly, instead of using lambdas.

vortexasdk/endpoints/vessel_movements_result.py
```diff
@@ -0,0 +1,361 @@
...
+
+        """
+        if columns is None:
+            columns = DEFAULT_COLUMNS
+
+        with Pool(os.cpu_count()) as pool:
+            records = pool.map(
+                functools.partial(
+                    convert_vessel_movement_to_flat_dict, cols=columns
+                ),
```
cvonsteg: again, might be cleaner to define the `partial` outside the context manager

PR #97, opened 2019-11-28
fix: Retry API calls on 500 status codes, fixes #95

### LINKED ISSUE #95: Make client more robust
https://circleci.com/gh/V0RT3X4/python-sdk/348

KitBurgess:


vortexasdk/retry_session.py
```diff
@@ -0,0 +1,24 @@
+from requests import Session
+from requests.adapters import HTTPAdapter
+from urllib3 import Retry
+
+
+# Inspired by https://www.peterbe.com/plog/best-practice-with-retries-with-requests
+def requests_retry_session(
+    retries=3,
+    backoff_factor=0.3,
+    status_forcelist=(500, 502, 504),
```
kujon: Let's not retry on `500` as this generally indicates there is a bug in the API. Retrying is simply going to trigger the bug again.

KitBurgess: Removed 500 from the list.

PR #119, opened 2019-12-04
Load all

### LINKED ISSUE #81: Add load all reference data method

KitBurgess:
This PR let's you do this, for each endpoint:
```
df = CargoMovements().load_all().to_df()
```

Closes #81

### DIFF
--- a/vortexasdk/config.py
+++ b/vortexasdk/config.py
@@ -1,4 +1,9 @@
 import os
+from datetime import datetime
+from dateutil.relativedelta import relativedelta
 
-LOG_LEVEL = os.getenv("LOG_LEVEL", "INFO")
+LOG_LEVEL = os.getenv("LOG_LEVEL", "DEBUG")
 LOG_FILE = os.getenv("LOG_FILE", "vortexasdk.log")
+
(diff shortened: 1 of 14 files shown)


cvonsteg:
Ah wait up - cliked on the wrong link, so my review only addressed one commit.  I'll reveiw the rest now

vortexasdk/config.py
```diff
@@ -1,4 +1,9 @@
 import os
+from datetime import datetime
+from dateutil.relativedelta import relativedelta
 
 LOG_LEVEL = os.getenv("LOG_LEVEL", "INFO")
 LOG_FILE = os.getenv("LOG_FILE", "vortexasdk.log")
+
+END_OF_AVAILABLE_DATA = datetime.utcnow() + relativedelta(months=6)
+BEGINNING_OF_AVAILABLE_DATA = END_OF_AVAILABLE_DATA - relativedelta(years=4)
```
cvonsteg: Are we time-boxing the data available to be loaded, even if clients use `load_all`?

KitBurgess: Yep, the API limits requests to 4 years max.

Ahh we could actually loop through batches of the min and max allowed date from the API

PR #131, opened 2019-12-09
perf: Shuffle offsets inplace for more accurate progress ETA, increas…

KitBurgess:
…e to 6threads


cvonsteg:
What is the benefit behind shuffling the list of offsets?

KitBurgess:
> What is the benefit behind shuffling the list of offsets?

Higher offset queries are slower than lower offset queries. If we loop sequentially through the offsets, in increasing order, then the progress bar underestimates the time required to loop through the remaining offsets.

Shuffling the offsets eradicates this problemo

PR #134, opened 2019-12-10
Remove fuzzy search

KitBurgess:
Enforce strict searching by IDs, removing the problem where a user would search for "India", and receive "Indiana", "British Indian Territories", "India"... etc.

We'll reintroduce fuzzy searching in a different PR under a different method (`fuzzy_search`) that can be used for exploratory analysis.

### DIFF
--- a/vortexasdk/api/shared_types.py
+++ b/vortexasdk/api/shared_types.py
@@ -58,7 +58,7 @@ class IDNameLayer:
     """Triple holding `id`, `name`, and `layer`."""
 
     id: ID
-    layer: str
+    layer: List[str]
     name: str
 
 
--- a/docs/examples/1_china.py
+++ b/docs/examples/1_china.py
@@ -24,7 +24,7 @@ china = [
 ]
 
 # Find the ID of all VLCCs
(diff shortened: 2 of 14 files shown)


vortexasdk/api/shared_types.py
```diff
@@ -58,7 +58,7 @@ class IDNameLayer:
     """Triple holding `id`, `name`, and `layer`."""
 
     id: ID
-    layer: str
+    layer: List[str]
```
KitBurgess: This is not related to PR, it's a bug fix

kujon:
Do we need to support `Union[ID, List[ID]]` for our filters? There isn't that much overhead in wrapping things in `[]`, but the code and the types are more complex as a result.

KitBurgess:
> Do we need to support `Union[ID, List[ID]]` for our filters? There isn't that much overhead in wrapping things in `[]` when querying, but the code and the types are more complex as a result.

If we were to only support `List[ID]`, we'd still have to check that the argument was indeed a `List[ID]`, instead of a plain old `ID`. The additional complexity of doing so is no less greater than wrapping an ID in a list.

PR #135, opened 2019-12-10
fix: search on vessel class is case insensitive

Natday:


tests/endpoints/test_vessels_real.py
```diff
@@ -68,3 +68,15 @@ def test_search_load_all_vessels(self):
...
+
+    def test_search_is_case_agnostic(self):
+        uppercase = {
+            v.id for v in Vessels().search(vessel_classes="Suezmax").to_list()
+        }
+        lowercase = {
+            v.id for v in Vessels().search(vessel_classes="suezmax").to_list()
+        }
+
+        two = 1 + 1
```
KitBurgess: We forgot to remove the mathematics

PR #154, opened 2020-02-21
feat: Product class now has hierarchy attribute

KitBurgess:


vortexasdk/api/product.py
```diff
@@ -15,6 +15,7 @@ class Product(Node, FromDictMixin):
 
     layer: List[str]
     meta: dict
+    hierarchy: List[Entity]
```
KitBurgess: This is the key change.

PR #158, opened 2020-03-03
feat: Check the number of records returned by API is what we'd expect

KitBurgess:


vortexasdk/client.py
```diff
@@ -75,6 +75,11 @@ def search(self, resource: str, **data) -> List:
 
         flattened = [x for y in responses for x in y]
 
+        assert len(flattened) == total, (
+            f"Incorrect number of records returned from API. "
```
ChloeConnor: maybe suggest re running their query in case it's not obvious?

PR #159, opened 2020-03-05
chore: Set Content-Type as application/json in requests

KitBurgess:


vortexasdk/retry_session.py
```diff
@@ -31,11 +31,14 @@ def _requests_retry_session(
     return session
 
 
+_headers = {"Content-Type": "application/json"}
```
cvonsteg: `nit`: If this is a global constant, convention is to set it as `_HEADERS` at the top of the file

PR #169, opened 2020-04-15
examples: add notebook with China Flows analysis

asfakianakis:
Added in the examples section the notebook with the China Flows analysis.


KitBurgess:
Should we run the SDK queries with less logging so that the notebook is even cleaner? https://vortechsa.github.io/python-sdk/config/

asfakianakis:
> Should we run the SDK queries with less logging so that the notebook is even cleaner? https://vortechsa.github.io/python-sdk/config/

Hmm, that's a good question. I am not sure what's better.

- In the one hand, I think that those detailed logs (especially when you are running live the notebook) give a sense of good anticipation to the user, i.e. it sees how many results are about to be retrieved, that the POST requests are send etc etc. So he feels like, "Hey this thing is working and it's just a matter of time until I have the data". I find this particularly useful in big requests because when you get a sense when you will have the data ready. I also think the logs give a more 'techy' mood. The SDK logs are well formatted and clear, so that's sth that probably would be appreciated by the more technical guys. I personally like to see the logs, but that's just my taste.

- On the other hand, it's true that some people might find this tiring or extraneous and probably the notebook would be more clear.

What if I add a comment using your link and saying if someone wants to disable the logs, here's how to do this.

asfakianakis:
> > Should we run the SDK queries with less logging so that the notebook is even cleaner? https://vortechsa.github.io/python-sdk/config/
> 
> Hmm, that's a very good question. I am not sure what's better.
> 
>     * In the one hand, I think that those detailed logs (especially when you are running live the notebook) give a sense of good anticipation to the user, i.e. it sees how many results are about to be retrieved, that the POST requests are send etc etc. So he feels like, "Hey this thing is working and it's just a matter of time until I have the data". I find this particularly useful in big requests because when you get a sense when you will have the data ready. I also think the logs give a more 'techy' mood. The SDK logs are well formatted and clear, so that's sth that probably would be appreciated by the more technical guys. I personally like to see the logs, but that's just my taste.
> 
>     * On the other hand, it's true that some people might find this tiring or extraneous and probably the notebook would be more clear.
> 
> 
> What if I add a comment using your link and saying if someone wants to disable the logs, here's how to do this.

Added this comment:

log_level_comment

PR #170, opened 2020-04-16
Feat/vm add ves status filter

asfakianakis:
Add `vessel_status` filter in the `VesselMovements` endpoint. This feature was a follow-up after the following discussion: https://vortexa.slack.com/archives/C8HLP5X8E/p1586920830292000.


docs/examples/4_ballast_movements.py
```diff
@@ -10,12 +10,8 @@
     search_result = VesselMovements().search(
         filter_time_min=datetime(2017, 10, 1, 0),
         filter_time_max=datetime(2017, 10, 1, 1),
+        filter_vessel_status="vessel_status_ballast",
```
asfakianakis, line 13: Just to incorporate the new filter instead of filtering on product retrospectively.

tests/endpoints/test_vessel_movements_vessel_status_filter.py
```diff
@@ -0,0 +1,16 @@
...
+class TestVesselMovementsRealVesselStatus(TestCaseUsingRealAPI):
+    def test_search(self):
+        v = VesselMovements().search(
+            filter_time_min=datetime(2019, 10, 1),
+            filter_time_max=datetime(2019, 10, 15),
+            filter_vessel_classes=["vlcc"],
+            filter_vessel_status="vessel_status_ballast",
+        )
+
+        assert len(v) > 50
```
asfakianakis, line 16: Maybe it's an overkill to add a whole test for a single filter? In any case, the reason for choosing 50 movements as a cutoff point was after the following inspection:

vessel_status_num_check

vortexasdk/endpoints/vessel_movements.py
```diff
@@ -42,6 +42,7 @@ def search(
         filter_products: Union[str, List[str]] = None,
         filter_vessels: Union[str, List[str]] = None,
         filter_vessel_classes: Union[str, List[str]] = None,
+        filter_vessel_status: str = "vessel_status_ballast",
```
KitBurgess, line 45: I think the default argument here should be None rather than defaulting to "vessel_status_ballast".

We can also annotate the type here.

tests/endpoints/test_vessel_movements_vessel_status_filter.py
KitBurgess: I agree 👍 - we could add this test method to the test_vessel_movements class, instead of adding a new test class.

docs/examples/4_ballast_movements.py
```diff
@@ -10,12 +10,8 @@
...
     )
 
     # Convert the search result to a dataframe
-    vessel_movements = search_result.to_df()
-
-    # Ballast movements are movements without a cargo.
-    is_ballast_mask = vessel_movements["cargoes.0.product.group.label"].isna()
-
-    ballast_movements = vessel_movements[is_ballast_mask]
```
KitBurgess, line 21: Nicee

vortexasdk/endpoints/vessel_movements.py
asfakianakis: Yes yes of course should be `None`. Will correct

tests/endpoints/test_vessel_movements_vessel_status_filter.py
asfakianakis: Done.

vortexasdk/endpoints/vessel_movements.py
asfakianakis: Done (I am not sure though about the annotation you mentioned).

PR #173, opened 2020-04-17
Update CONTRIBUTING.md

KitBurgess:


CONTRIBUTING.md
```diff
@@ -75,6 +75,9 @@ A good pull request:
 Tips:
 - If you're adding a new endpoint, adapt and use `./docs/generate_stubs.sh` to generate sample json data used for tests.
 
+### Contributing Jupyter Notebooks
+
+Please include the version of the SDK used to generate the Jupyter Notebook at the start of the notebook. This helps others to reproduce your results in the future.
```
asfakianakis, line 80: Really good point!! Should we also mention which other additional packages were installed (i.e. the ones that are not part of SDK requirements)?

Also next step is to create a Jupyter Notebook template and from now on all notebooks follow the same template.

PR #177, opened 2020-04-22
chore: Log the actual payload sent to the API

KitBurgess:


vortexasdk/operations.py
```diff
@@ -79,7 +79,5 @@ def search(self, **params) -> List[dict]:
         >>> Search("/reference/vessels").search(term="DHT") # doctest: +SKIP
 
         """
-        logger.info(
```
KitBurgess, line 82: Previously we logged the payload before elements were filtered. This was  not helpful - as we weren't actually logging the payload sent to the API

vortexasdk/client.py
```diff
@@ -40,7 +41,9 @@ def get_reference(self, resource: str, id: ID) -> List[Dict]:
     def search(self, resource: str, **data) -> List:
         """Search using `resource` using `**data` as filter params."""
         url = self._create_url(resource)
-        payload = {k: v for k, v in data.items() if v is not None}
+        payload = self._cleanse_payload(data)
+        logger.info(f"Payload: {payload}")
```
KitBurgess, line 45: Now we log the actual payload sent to the API

vortexasdk/client.py
```diff
@@ -95,7 +98,14 @@ def _process_multiple_pages(
                 return pool.map(func, offsets)
 
     @staticmethod
-    def _calculate_total(response) -> int:
+    def _cleanse_payload(payload: Dict) -> Dict:
+        filtered = {
```
KitBurgess, line 102: We filter this payload.

It's also sorted for consistency when reading the logs.

PR #178, opened 2020-04-27
Feat/origin dest exclusion filters

asfakianakis:
Add exclusion filters to `CargoMovements` and `VesselMovements` endpoints. Closes [#168 ](https://github.com/VorTECHsa/python-sdk/issues/168) and on top includes all possible 'exclude' filters (e..g. excluce charterers, owners, products etc.)

### DIFF
--- a/tests/endpoints/test_cargo_movements_real.py
+++ b/tests/endpoints/test_cargo_movements_real.py
@@ -3,7 +3,7 @@ from datetime import datetime
 from docs.utils import to_markdown
 from tests.testcases import TestCaseUsingRealAPI
 from tests.timer import Timer
-from vortexasdk import Geographies, Corporations
+from vortexasdk import Geographies, Corporations, Products
 from vortexasdk.endpoints.cargo_movements import CargoMovements
 
 
@@ -46,6 +46,71 @@ class TestCargoMovementsReal(TestCaseUsingRealAPI):
         )
 
         assert len(df) == 2
+
+    def test_exlusion_filter(self):
+        crude = [
+            p.id
+            for p in Products().search("Crude/Condensates").to_list()
+            if p.layer == ["group"]
+        ]
+        arab_medium = [
+            p.id
+            for p in Products().search("Arab Medium").to_list()
+            if p.layer == ["grade"]
+        ]
+        meg = [
+            g.id
+            for g in Geographies().search("MEG/AG").to_list()
+            if "trading_region" in g.layer
+        ]
+        iraq = [
+            g.id
+            for g in Geographies().search("Iraq").to_list()
+            if "country" in g.layer
+        ]
+
+        cols = [
+            "cargo_movement_id",
+            "vessels.0.name",
+            "events.cargo_port_load_event.0.location.port.id",
+            "events.cargo_port_load_event.0.location.port.label",
+            "events.cargo_port_load_event.0.location.country.id",
+            "events.cargo_port_load_event.0.location.country.label",
+            "events.cargo_port_load_event.0.start_timestamp",
+            "events.cargo_port_load_event.0.end_timestamp",
+            "product.grade.id",
+            "product.grade.label",
+            "product.group_product.label",
+        ]
+
+        df = (
+            CargoMovements()
+            .search(
+                filter_activity="loading_end",
+                filter_products=[
+                    "54af755a090118dcf9b0724c9a4e9f14745c26165385ffa7f1445bc768f06f11"
+                ],
+                filter_origins=meg,
+                exclude={
+                    "filter_origins": iraq,
+                    "filter_products": arab_medium,
+                },
+                filter_time_min=datetime(2019, 10, 1),
+                filter_time_max=datetime(2019, 11, 1),
+                cm_unit="b",
+            )
+            .to_df(columns=cols)
+        )
+
+        df_excl = df.loc[
+            (
+                df["events.cargo_port_load_event.0.location.country.id"]
+                == iraq[0]
+            )
+            | (df["product.grade.id"] == arab_medium[0])
+        ]
+
+        assert df_excl.empty
 
     def test_to_df_all_columns(self):
         df = (
--- a/tests/endpoints/test_cargo_movements_real.py
+++ b/tests/endpoints/test_cargo_movements_real.py
@@ -92,7 +92,7 @@ class TestCargoMovementsReal(TestCaseUsingRealAPI):
                 ],
                 filter_origins=meg,
                 exclude={
-                    "filter_origins": iraq,
+                    "filter_origins": iraq[0],
                     "filter_products": arab_medium,
                 },
                 filter_time_min=datetime(2019, 10, 1),
--- a/tests/endpoints/test_cargo_movements_real.py
+++ b/tests/endpoints/test_cargo_movements_real.py
@@ -91,10 +91,8 @@ class TestCargoMovementsReal(TestCaseUsingRealAPI):
                     "54af755a090118dcf9b0724c9a4e9f14745c26165385ffa7f1445bc768f06f11"
                 ],
                 filter_origins=meg,
-                exclude={
-                    "filter_origins": iraq[0],
-                    "filter_products": arab_medium,
-                },
+                exclude_origins=iraq[0],
+                exclude_products=arab_medium,
                 filter_time_min=datetime(2019, 10, 1),
                 filter_time_max=datetime(2019, 11, 1),
                 cm_unit="b",
--- a/tests/endpoints/test_vessel_movements_real.py
+++ b/tests/endpoints/test_vessel_movements_real.py
@@ -61,7 +61,10 @@ class TestVesselMovementsReal(TestCaseUsingRealAPI):
             VesselMovements()
             .search(
                 filter_origins=meg,
-                exclude={"filter_origins": iraq, "filter_charterers": bahri},
+                exclude={
+                    "filter_origins": iraq,
+                    "filter_charterers": bahri[0],
+                },
                 filter_time_min=datetime(2019, 10, 15),
                 filter_time_max=datetime(2019, 11, 1),
             )
--- a/tests/endpoints/test_vessel_movements_real.py
+++ b/tests/endpoints/test_vessel_movements_real.py
@@ -1,7 +1,7 @@
 from datetime import datetime
 
 from tests.testcases import TestCaseUsingRealAPI
-from vortexasdk import VesselMovements, Geographies
+from vortexasdk import VesselMovements, Geographies, Corporations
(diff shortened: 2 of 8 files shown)


tests/endpoints/test_cargo_movements_real.py
```diff
@@ -47,6 +47,71 @@ def test_search_single_filter_id(self):
...
+
+        df_excl = df.loc[
+            (
+                df["events.cargo_port_load_event.0.location.country.id"]
+                == iraq[0]
+            )
+            | (df["product.grade.id"] == arab_medium[0])
+        ]
+
+        assert df_excl.empty
```
asfakianakis, line 113: test_cms_excl

tests/endpoints/test_vessel_movements_real.py
```diff
@@ -30,6 +30,51 @@ def test_search_vessel_status(self):
...
+            .to_df(columns=cols)
+        )
+
+        mask = (df["origin.location.country.id"] == iraq[0]) | (
+            df["vessel.corporate_entities.charterer.id"] == bahri[0]
+        )
+        df_excl = df.loc[mask]
+
+        assert df_excl.empty
+
```
asfakianakis, line 77: test_vm_excl

vortexasdk/endpoints/cargo_movements.py
```diff
@@ -40,6 +40,7 @@ def search(
         filter_storage_locations: Union[ID, List[ID]] = None,
         filter_ship_to_ship_locations: Union[ID, List[ID]] = None,
         filter_waypoints: Union[ID, List[ID]] = None,
+        exclude: Dict[ID, Union[ID, List[ID]]] = None,
```
asfakianakis, line 43: Actually that's the most important change. Exclusion filter must be a `dict` where keys are the parameters we want to apply the exlusions (e.g. `filter_products`, `filter_origins` etc.) and values are the corresponding IDs. I attach an image to visualize a relevant POST request in the UI.

exclude_query_POST

vortexasdk/endpoints/cargo_movements.py
```diff
@@ -78,6 +79,10 @@ def search(
 
             filter_waypoints: A geography ID, or list of geography IDs to filter on.
 
+            exclude: A dictionary where keys are strings that represent the parameter to implement exclusions on and
+                values are IDs or list of IDs that correspond to the entities to be exluded. Dictionary keys must be one of
+                ['filter_origins', 'filter_destinations', 'filter_products', 'filter_vessels', 'filter_charterers', 'filter_owners']
+
```
asfakianakis, line 85: Is that clear enough? Should we maybe add or modify any current example to illustrate the exact use of this filter?

vortexasdk/endpoints/cargo_movements.py
```diff
@@ -178,6 +183,7 @@ def search(
                 filter_ship_to_ship_locations
             ),
             "filter_waypoints": convert_to_list(filter_waypoints),
+            "exclude": convert_values_to_list_abstract(exclude),
```
asfakianakis, line 186: Same logic as in `CargoMovements` endpoint.

vortexasdk/utils.py
```diff
@@ -14,3 +14,14 @@ def convert_to_list(a) -> List:
...
+
+def convert_values_to_list_abstract(a) -> Dict:
+    """ If element is None returns empty dict else converts each value to list if it isn't a list already."""
+    if isinstance(a, dict):
+        for k, v in a.items():
+            if not isinstance(v, list):
+                a[k] = convert_to_list(v)
+        return a
+    else:
+        return {}
```
asfakianakis, line 27: This function is applied to the exclusion argument and converts the values to list if necessary.

vortexasdk/endpoints/vessel_movements.py
```diff
@@ -43,6 +43,7 @@ def search(
         filter_vessels: Union[str, List[str]] = None,
         filter_vessel_classes: Union[str, List[str]] = None,
         filter_vessel_status: str = None,
+        exclude: Dict[str, Union[str, List[str]]] = None,
```
asfakianakis, line 46: Same logic as in the `CargoMovements` endpoint.

tests/endpoints/test_cargo_movements_real.py
```diff
@@ -47,6 +47,71 @@ def test_search_single_filter_id(self):
...
+            .search(
+                filter_activity="loading_end",
+                filter_products=[
+                    "54af755a090118dcf9b0724c9a4e9f14745c26165385ffa7f1445bc768f06f11"
+                ],
+                filter_origins=meg,
+                exclude={
+                    "filter_origins": iraq[0],
+                    "filter_products": arab_medium,
+                },
```
KitBurgess, line 97: I think here we could make this even simpler and present `exclude_product = arab_medium` as a top level argument to the user. Creating a dictionary of exclusions is an API implementation detail, and something we can take care of internally in the SDK.

tests/endpoints/test_vessel_movements_real.py
```diff
@@ -30,6 +30,54 @@ def test_search_vessel_status(self):
...
+                    "filter_origins": iraq,
+                    "filter_charterers": bahri[0],
+                },
+                filter_time_min=datetime(2019, 10, 15),
+                filter_time_max=datetime(2019, 11, 1),
+            )
+            .to_df(columns=cols)
+        )
+
+        mask = (df["origin.location.country.id"] == iraq[0]) | (
```
KitBurgess, line 74: Like these tests 👍

vortexasdk/endpoints/cargo_movements.py
```diff
@@ -60,7 +61,7 @@ def search(
 
             cm_unit: Unit of measurement. Enter 'b' for barrels or 't' for tonnes.
 
-            filter_corporations: A corporation ID, or list of corporation IDs to filter on.
+            filter_charterers: A charterer ID, or list of charterer IDs to filter on.
```
KitBurgess, line 64: Nice bugfix!

vortexasdk/endpoints/cargo_movements.py
KitBurgess: Similarly we can have:
```
CargoMovements.search(filter_origins=us, exclude_origins=padd1])
```

this keeps it simple and explicit for the user, avoiding nesting.

tests/endpoints/test_cargo_movements_real.py
asfakianakis: That's a really good point, you are right. Will try this.

asfakianakis:
```
def wrap_args_to_dict(foo, keyword='exclude') -> Dict:
    """
    Paremeters
    ----------
        foo: function
        keyword: keeps only arguments that contain this keyword
    """
    # Store function args and deafault values
    obj = inspect.getfullargspec(foo)
    din = dict(zip(obj.args, obj.defaults))
    
    # Filter dict based on keyword
    d = dict()
    for k,v in din.items():
        if keyword in k:
            d[k] = convert_to_list(v)
            
    return d
    ```

Comment (asfakianakis) [reply tests/endpoints/test_cargo_movements_real.py:97]: @KitBurgess Before I make any commits, just for some feedback, does the following implementation look fine?
```
def wrap_args_to_dict(foo, keyword='exclude') -> Dict:
    """
    Paremeters
    ----------
        foo: function
        keyword: keeps only arguments that contain this keyword
    """
    # Store function args and deafault values
    obj = inspect.getfullargspec(foo)
    din = dict(zip(obj.args, obj.defaults))
    
    # Filter dict based on keyword
    d = dict()
    for k,v in din.items():
        if keyword in k:
            d[k] = convert_to_list(v)
            
    return d
```

So if we have a function `foo` like this:

```
def foo(exclude_origins='3', exclude_charterers='4', exclude_owners=None, c=5):
    obj = inspect.getfullargspec(test)
    d = dict(zip(obj.args, obj.defaults))
    return d
```

When you call `wrap_args_to_dict(foo,'exclude')` you would get:

```
{'exclude_origins': ['3'], 'exclude_charterers': ['4'], 'exclude_owners': []}
```

So you would then go to `CargoMovements` class and in the `params` use:

```
"exclude": wrap_args_to_dict(search)
```

Comment (KitBurgess) [reply tests/endpoints/test_cargo_movements_real.py:97]: Sorry - I'm not sure I follow this, it looks more complicated though than explicitly creating the `exclude` dictionary, and then filtering this / wrapping the values in a dict.

I think we should commit a working simple solution first, then look to reduce any duplicated code afterwards.

Comment (asfakianakis) [reply tests/endpoints/test_cargo_movements_real.py:97]: The `wrap_args_to_dict` function does the following: It finds all the arguments and default values of a given input function, it then creates a dict where the dict keys are the function arguments and the dict values the default arg values and finally filters the dict by keeping all the pairs where the keys include the specified keyword. 

In our case when we call this function inside the `CargoMovements` class like wrap_args_to_dict(search), it will create the corresponding dict with all arguments that include 'exclude'.

Of course, the other way as you mention is to just define an exclusion dictionary before define the params dict. We would just need to do this in every class, and if for some reasons an endpoint has different arguments than another (e.g. `VesselMovements` has the `filter_vessel_class` filter already deployed, whereas `CargoMovements` not) it would be more difficult I think to maintain.

File: vortexasdk/endpoints/cargo_movements.py [hunk 11]
```diff
@@ -40,6 +40,12 @@ def search(
         filter_storage_locations: Union[ID, List[ID]] = None,
         filter_ship_to_ship_locations: Union[ID, List[ID]] = None,
         filter_waypoints: Union[ID, List[ID]] = None,
+        exclude_origins: Union[ID, List[ID]] = None,
+        exclude_destinations: Union[ID, List[ID]] = None,
+        exclude_products: Union[ID, List[ID]] = None,
+        exclude_vessels: Union[ID, List[ID]] = None,
+        exclude_charterers: Union[ID, List[ID]] = None,
+        exclude_owners: Union[ID, List[ID]] = None,
```
asfakianakis, line 48: Simplify exclusion filters by adding as separate arguments.

vortexasdk/endpoints/cargo_movements.py
```diff
@@ -158,6 +176,15 @@ def search(
...
 
         """
+        exclude_params = {
+            "filter_origins": convert_to_list(exclude_origins),
+            "filter_destinations": convert_to_list(exclude_destinations),
+            "filter_products": convert_to_list(exclude_products),
+            "filter_vessels": convert_to_list(exclude_vessels),
+            "filter_charterers": convert_to_list(exclude_charterers),
+            "filter_owners": convert_to_list(exclude_owners),
+        }
```
asfakianakis, line 186: Explicitly define exclusion dictionary.

PR #182, opened 2020-05-01
feat: Add version to API params

### LINKED ISSUE #179: Add python SDK version to API calls
From @JRGranell :

Add `_sdk=python_vxxx` (where xxx is the current SDK version) to the URL query string? So when a URL is generated it might look like https://api.vortexa/com/v5/search/cargo-movements/?apikey=xxx&_sdk=python_v1.3.2 this will allow us to track which API calls come from which sources and SDKs (we have a JS one) which will help us track usage, and identify issues quicker.

KitBurgess:
Closes #179


KitBurgess:
Following the advice here: https://packaging.python.org/guides/single-sourcing-package-version/

setup.py
```diff
@@ -3,9 +3,13 @@
 with open("README.md", "r") as fh:
     long_description = fh.read()
 
+version = {}
+with open("./vortexasdk/version.py") as fp:
+    exec(fp.read(), version)
```
cvonsteg, line 8: could you replace this with `from vortexasdk.version import __version__`?

cvonsteg: Not 100% sure whether this works in a setup.py script

vortexasdk/client.py
```diff
@@ -67,7 +68,9 @@ def search(self, resource: str, **data) -> List:
             return flattened
 
     def _create_url(self, path: str) -> str:
-        return f"{API_URL}{path}?apikey={self.api_key}"
+        return (
+            f"{API_URL}{path}?_sdk=python_v{__version__}&apikey={self.api_key}"
```
cvonsteg, line 72: 👍

setup.py
KitBurgess: So interestingly we can't do this (I tried). 

The reason we can't is because when executes `setup.py` in order to create the `vortexasdk` module. pip hasn't gathered or installed the requirements yet, so attempting to `import vortexasdk.version` crashes.

cvonsteg: Nice, I figured you might have tried already.  Good to know.

PR #185, opened 2020-05-04
fix: Remove exclude empties from payload

KitBurgess:


vortexasdk/client.py
```diff
@@ -98,10 +99,10 @@ def _process_multiple_pages(
...
     @staticmethod
     def _cleanse_payload(payload: Dict) -> Dict:
-        filtered = {
-            k: v for k, v in payload.items() if not (v is None or v == [])
-        }
-        return dict(sorted(filtered.items()))
+        exclude_params = payload.get("exclude", {})
+        payload["exclude"] = filter_empty_values(exclude_params)
+
+        return filter_empty_values(payload)
```
asfakianakis, line 105: Nice abstraction with `filter_empty_values`.

tests/test_client.py
```diff
@@ -9,4 +9,16 @@ def test__cleanse_payload(self):
...
+        payload = {
+            "layer1_k1": [],
+            "layer1_k2": None,
+            "layer1_k3": 3,
+            "exclude": {"layer2_k1": [], "layer2_k2": [1]},
+        }
+
+        cleansed = VortexaClient._cleanse_payload(payload)
+
+        assert cleansed == {"layer1_k3": 3, "exclude": {"layer2_k2": [1]}}
```
asfakianakis, line 24: Nice test as well!

vortexasdk/client.py
asfakianakis: One question actually: What happens if  there is no `exclude` field in the payload, e.g. when we query `Geographies/Vessels/Products` endpoints? Should we add a check?

KitBurgess: the `payload.get("exclude", {})` is a "get or default" method ie it gets the value for the `"exclude"` key if it exists, else returns `{}`.

KitBurgess: You're right to say we should test it - this test (implicitly) runs that logic, because there's no `"exclude"` key in the dict

https://github.com/VorTECHsa/python-sdk/blob/fix/remove_empties_payload/tests/test_client.py#L7

asfakianakis: Understood. I was referring to tests actually, but was confused because I thought that the payload will contain an `exclude` parameter with an empty dict as value, so a POST request e.g. to `Vessels` endpoint might fail. But `filter_empty_values` takes care of it and removes the empty values. All good then , thanks for the explanation.

PR #190, opened 2020-05-05
Feat attribute endpoint 

### LINKED ISSUE #189: Add attributes reference endpoint
The API has recently had a new reference endpoint added to it: `/reference/attributes`

This needs to be added to the SDK.

SeanBarry:
This PR adds the ability to query the new reference/attributes endpoint in the API. It also adds the necessary tests.

This PR closes #189

### DIFF
--- a/vortexasdk/endpoints/attributes.py
+++ b/vortexasdk/endpoints/attributes.py
@@ -7,16 +7,51 @@ from vortexasdk.utils import convert_to_list
 
 
 class Attributes(Reference, Search):
-    """Attributes endpoint."""
+    """
+        Attributes endpoint.
+
+        An Attribute is a reference value that corresponds to an ID associated with other entities.
+
+        For example, a vessel object from the Vessel reference endpoint may have the following keys:
+
+        ```json
+        {
+            "ice_class": "b09ed4e2bd6904dd",
+            "propulsion": "3ace0e050724707b"
+        }
+        ```
+
+        These IDs represent attributes which can be found via the Attributes reference endpoint.
+
+        When the attributes endpoint is searched with those ids as parameters:
+
+        ```python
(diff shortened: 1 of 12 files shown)


vortexasdk/endpoints/attributes.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def __init__(self):
+        Reference.__init__(self, ATTRIBUTES_REFERENCE)
+        Search.__init__(self, ATTRIBUTES_REFERENCE)
+
+    def load_all(self) -> AttributeResult:
+        """Load all attributes."""
+        return self.search()
+
+    def search(
```
KitBurgess, line 20: We're shadowing the inbuilt python keyword `type` here.

I don't think it's too big a problem to do so here, but we can annotate the code to alert any linters / code quality checkers that we're aware of this.

```suggestion
	# noinspection PyShadowingBuiltins
    def search(
```

KitBurgess:
Before we merge this PR, should we branch off of this, then create a separate PR to close #191  -> that way we can merge the two as part of the same release.

This means users will immediately be able to use the Attributes functionality. It might also prevent confusion, as users will no doubt be thinking _"This Attributes endpoint is 🔥 , but what can I use it for!"_

KitBurgess:
@all-contributors add @SeanBarry for code, doc

SeanBarry:
@KitBurgess 

> The only thing I'd add, is that it'd be good to expand on what attributes are - and to give some context to them. (I assume they're vessel attributes).
Good suggestion, I've added this.

> On other thing, the example shows searching for all attributes of type="ice_class", but the dataframe returned shows two attributes of type scrubber
Good spot, this was a mistake on my part. I've fixed it

PR #193, opened 2020-05-05
Feat new filters

### LINKED ISSUE #191: Filter vessel movements and cargo movements with attributes, age, flag & scrubbers.
Prerequisite: #189

SeanBarry:
The API recently added some new filters

Vessel reference: 
- scrubbers

Cargo movements/Vessel movements
- scrubbers
- min vessel age
- max vessel age
- vessel flags (inc/exc)
- vessel ice class (inc/exc)
- vessel propulsion (inc/exc)

The underlying shape of the responses was changed slightly too.
vessel entities now have the following fields:

`year`, `scrubber`, `flag`

Closes #191


vortexasdk/enums.py
```diff
@@ -0,0 +1,12 @@
...
+
+class VesselAttributesEnum(str, Enum):
+    FlagTag = "vessel_flag_tag"
+    ScrubberTag = "vessel_scrubber_tag"
+
+
+class ScrubbersFittedEnum(str, Enum):
+    Off = ("off",)
+    Included = ("inc",)
+    Excluded = "exc"
```
KitBurgess, line 12: I love these enums, but I sadly I don' think we can use them.

Enums mean that the SDK is dependent on the values returned by the  Vortexa API, which sits outside of any semver schema changes.

For example, the API could return a new ScrubbersFitted option, `foo`, without any major, minor, or even patch semver change. This would raise an exception in the SDK when the SDK attempts to serialize `foo` into the `ScrubbersFittedEnum`. For this reason, we're safer off sticking to strings in cases like these 👍

KitBurgess: ^^ The above statement is actually incorrect. Python can serialize a string into an non-existing enum.

KitBurgess:
@cvonsteg @kujon what do you think of using enums in the SDK. I'm on the fence, slash leaning towards thinking we shouldn't.

Enums pros and cons

Pros:

- better autocomplete
- clearer codebase for contributors
- shows users some potential values for arguments

Cons:

- Enums couple the SDK with API
- Enums will easily get out of date, which renders them less-than-helpful, because they may contain values that aren’t in the API, and the API may contain values that aren’t in the enum

cvonsteg:
@KitBurgess @kujon - I do like Enums, but in short, I think it may be best to avoid them in this case.

The reasons are:
1) As you rightly pointed out @KitBurgess , using enums the way we're proposing here, means the SDK becomes more tightly coupled to the API.  This means there is an onus on us to always keep them up-to-date.  The risk is that we announce updates to the API to our clients, but fall behind, or forget to update the SDK.  When the 2 are out-of-sync it looks bad on us as we cannot provide the full functionality via the SDK, or create a barrier for clients to move work that they started with direct api calls, into the SDK.  It would probably mean that ADT would have to ensure the SDK is up-to-date with any API changes.
2) Enums are less common in python than other languages.  Making them a core data structure may make the user-experience a bit more tedious or confusing for less experienced users.

kujon:
I don't see an issue tbh. This is just encoding what vessel attribute types you might get from this endpoint.

SeanBarry:
Users of the SDK will never use the enum. It's only ever developers working on the code. The way users interact with the SDK will be exactly the same as before.

KitBurgess:
Let's decouple using enums from this PR, that way we can get this PR out, giving clients the new vessel attribute functionality.

If we do go ahead with enums, then we'll want to introduce them in other places in the codebase, in a separate PR. I believe enums **will** make the SDK a better product, we just need to ensure they're kept up to date in a programmatic way.

Given that this current new-filter feature PR is very nearly ready to go, @SeanBarry can you cherrypick the enum work to a separate branch? We can keep this `feat-new-filters` branch using strings for now and get it out in the wild.

SeanBarry:
Sounds good to me, thanks for the input guys!

vortexasdk/endpoints/vessel_movements.py
```diff
@@ -36,21 +37,30 @@ def search(
...
-        filter_products: Union[str, List[str]] = None,
-        filter_vessels: Union[str, List[str]] = None,
-        filter_vessel_classes: Union[str, List[str]] = None,
+        filter_charterers: Union[ID, List[ID]] = None,
+        filter_destinations: Union[ID, List[ID]] = None,
+        filter_origins: Union[ID, List[ID]] = None,
+        filter_owners: Union[ID, List[ID]] = None,
+        filter_products: Union[ID, List[ID]] = None,
+        filter_vessels: Union[ID, List[ID]] = None,
+        filter_vessel_classes: Union[ID, List[ID]] = None,
```
KitBurgess, line 46: Nice change

vortexasdk/endpoints/vessel_movements.py
```diff
@@ -147,6 +182,16 @@ def search(
...
+            "filter_vessel_age_min": filter_vessel_age_min,
+            "filter_vessel_age_max": filter_vessel_age_max,
+            "filter_vessel_scrubbers": filter_vessel_scrubbers,
+            "filter_vessel_flags": convert_to_list(filter_vessel_flags),
+            "filter_vessel_ice_class": convert_to_list(
+                filter_vessel_ice_class
+            ),
+            "filter_vessel_propulsion": convert_to_list(
+                filter_vessel_propulsion
+            ),
```
KitBurgess, line 194: Should we add a simple test, calling the API with these filters?

PR #198, opened 2020-05-27
Feat/version notification

### LINKED ISSUE #126: Alert users of SDK if they're using an outdated version
We'll need a toggle in config to turn this on/off

asfakianakis:
Raise a warning when a user is using an old SDK version. The user would see something like this:

user_warning

This closes #126.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -23,7 +23,7 @@ class TestUtils(TestCase):
 
         assert convert_values_to_list(d) == expected
 
-    def test_get_latest_package_version(self):
+    def test_get_latest_sdk_version(self):
         sdk_old_version = "0.18.0"
 
         lv = get_latest_sdk_version()
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1,8 +1,10 @@
 from unittest import TestCase
+from distutils.version import StrictVersion
 
 from vortexasdk.utils import (
     convert_to_list,
     convert_values_to_list,
+    get_latest_package_version,
 )
 
 
@@ -20,3 +22,10 @@ class TestUtils(TestCase):
         expected = {1: [], 2: ["b"], 3: ["c"]}
 
         assert convert_values_to_list(d) == expected
+
+    def test_get_latest_package_version(self):
+        pd_old_version = "0.25.3"
+
+        lv, _ = get_latest_package_version("pandas")
+
+        assert lv > StrictVersion(pd_old_version)
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1,10 +1,10 @@
 from unittest import TestCase
-from distutils.version import StrictVersion
+from distutils.version import LooseVersion
 
 from vortexasdk.utils import (
     convert_to_list,
     convert_values_to_list,
-    get_latest_package_version,
+    get_latest_sdk_version,
 )
 
 
@@ -24,8 +24,8 @@ class TestUtils(TestCase):
         assert convert_values_to_list(d) == expected
 
     def test_get_latest_package_version(self):
-        pd_old_version = "0.25.3"
+        sdk_old_version = "0.18.0"
 
-        lv, _ = get_latest_package_version("pandas")
+        lv = get_latest_sdk_version()
 
-        assert lv > StrictVersion(pd_old_version)
+        assert lv > LooseVersion(sdk_old_version)
--- a/vortexasdk/client.py
+++ b/vortexasdk/client.py
@@ -17,8 +17,10 @@ from vortexasdk.retry_session import (
     retry_get,
     retry_post,
 )
-from vortexasdk.utils import filter_empty_values
+from vortexasdk.utils import filter_empty_values, get_latest_package_version
 from vortexasdk.version import __version__
+from vortexasdk import __name__ as pkg_name
+from distutils.version import StrictVersion, LooseVersion
 
 logger = get_logger(__name__)
 
@@ -194,6 +196,23 @@ def create_client() -> VortexaClient:
     except KeyError:
         raise KeyError(
             "VORTEXA_API_KEY environment variable is required to use the VortexaSDK"
+        )
+    latest_version, _ = get_latest_package_version(pkg_name)
+    try:
+        version_mask = StrictVersion(__version__) < StrictVersion(
+            latest_version
+        )
+    except ValueError as e:
+        if "invalid version number" in str(e):
+            version_mask = LooseVersion(__version__) < LooseVersion(
+                latest_version
+            )
+        else:
+            raise e
+    if version_mask:
+        logger.warning(
+            f"You are using vortexasdk version {__version__}, however version {latest_version} is available.\n"
+            f"You should consider upgrading via the 'pip install vortexasdk --upgrade' command."
         )
     return VortexaClient(api_key=api_key)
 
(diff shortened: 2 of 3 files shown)


vortexasdk/utils.py
```diff
@@ -20,3 +23,29 @@ def filter_empty_values(data: Dict) -> Dict:
     return {
         k: v for k, v in data.items() if not (v is None or v == [] or v == {})
     }
+
+
+def get_latest_package_version(package_name: str):
```
asfakianakis, line 28: That's the function that retrieves the latest SDK version from PyPI.

vortexasdk/client.py
```diff
@@ -195,6 +197,23 @@ def create_client() -> VortexaClient:
...
+            version_mask = LooseVersion(__version__) < LooseVersion(
+                latest_version
+            )
+        else:
+            raise e
+    if version_mask:
+        logger.warning(
+            f"You are using vortexasdk version {__version__}, however version {latest_version} is available.\n"
+            f"You should consider upgrading via the 'pip install vortexasdk --upgrade' command."
+        )
```
asfakianakis, line 216: I added the warning at the point where we are creating the client. Not sure if that's the best place to do so. Also open to any comments on whether we could do this code more robust, i.e. that's a 'nice-to-have' check, we don't want this to be the cause of failing to create the client.

tests/test_utils.py
```diff
@@ -20,3 +22,10 @@ def test_convert_values_to_list(self):
         expected = {1: [], 2: ["b"], 3: ["c"]}
 
         assert convert_values_to_list(d) == expected
+
+    def test_get_latest_package_version(self):
+        pd_old_version = "0.25.3"
+
+        lv, _ = get_latest_package_version("pandas")
+
+        assert lv > StrictVersion(pd_old_version)
```
asfakianakis, line 31: Test to make sure the package version function works as expected.

vortexasdk/utils.py
```diff
@@ -20,3 +23,29 @@ def filter_empty_values(data: Dict) -> Dict:
...
+        versions = sorted(versions, key=StrictVersion)
+    except ValueError as e:
+        if "invalid version number" in str(e):
+            versions = sorted(versions, key=LooseVersion)
+        else:
+            raise e
+
+    latest_version = versions[-1]
+
+    return latest_version, versions
```
asfakianakis, line 51: That's the function that retrieves the latest SDK version from PyPI.

vortexasdk/client.py
KitBurgess, line 216: I think this is a good place to put the logic 👍 .

Can we extract this into a separate function though, then just call it in a single line in the `create_client` method

vortexasdk/utils.py
KitBurgess: We only ever use `latest_version` here, so let's just return `latest_version`

KitBurgess: Then we can also annotate this method with `-> str:`

vortexasdk/client.py
KitBurgess: Any yes, that's a good idea. The entire content of the `warn_user_if_sdk_is_outdated` method can be wrapped in a try-catch method just in case it blows up, so as not to prevent the client from being used.

vortexasdk/utils.py
```diff
@@ -20,3 +23,29 @@ def filter_empty_values(data: Dict) -> Dict:
...
+    with urlopen(url) as u:
+        data = json.loads(u.read())
+
+    try:
+        versions = data["releases"].keys()
+    except KeyError:
+        raise KeyError("No version related key found in PyPI")
+
+    try:
+        versions = sorted(versions, key=StrictVersion)
```
KitBurgess, line 42: rather than mutate the `versions` variable here, we should create a new variable `sorted_versions`. We should generally write immutable variables when possible - it's easier to debug.

vortexasdk/client.py
```diff
@@ -195,6 +197,23 @@ def create_client() -> VortexaClient:
...
             "VORTEXA_API_KEY environment variable is required to use the VortexaSDK"
         )
+    latest_version, _ = get_latest_package_version(pkg_name)
+    try:
+        version_mask = StrictVersion(__version__) < StrictVersion(
+            latest_version
+        )
+    except ValueError as e:
+        if "invalid version number" in str(e):
+            version_mask = LooseVersion(__version__) < LooseVersion(
```
KitBurgess, line 207: I don't think the variable name `version_mask` is a mask, it's more a boolean representing `sdk_version_is_out_of_date`.

Given that our public versions all follow semver, I think we're safe to just use the `LooseVersion` equality check (and not need the `StrictVersion` try catch).

vortexasdk/utils.py
```diff
@@ -20,3 +23,29 @@ def filter_empty_values(data: Dict) -> Dict:
     return {
         k: v for k, v in data.items() if not (v is None or v == [] or v == {})
     }
+
+
+def get_latest_package_version(package_name: str):
+    """
+    Retrieves the latest SDK version from PyPI and compares with the currently installed SDK version.
+    """
```
KitBurgess, line 31: We currently only ever call this method with the argument `(from vortexasdk import __name__ as pkg_name)`, so we could do away with the argument, and call it `get_latest_sdk_version`. This makes the code simpler.

vortexasdk/client.py
asfakianakis: All of those are fair points. Will apply the requested changes and revert.

vortexasdk/client.py
```diff
@@ -205,3 +208,19 @@ def set_client(client) -> None:
...
+        if sdk_outdated_check:
+            logger.warning(
+                f"You are using {sdk_pkg_name} version {__version__}, however version {latest_sdk_version} is available.\n"
+                f"You should consider upgrading via the 'pip install {sdk_pkg_name} --upgrade' command."
+            )
+    except Exception as e:
+        logger.warning(
+            f"Outdated SDK version check could not be completed. \n"
+            f"Got an exception: {e}"
+        )
```
asfakianakis, line 226: Separate function in the client script to raise a warning if the SDK version is outdated.

vortexasdk/utils.py
```diff
@@ -20,3 +25,25 @@ def filter_empty_values(data: Dict) -> Dict:
...
+    return latest_version
+
+
+def is_sdk_version_outdated():
+    """Checks whether SDK version is outdated."""
+    latest_version = get_latest_sdk_version()
+    if LooseVersion(__version__) < latest_version:
+        return latest_version, True
+    else:
+        return latest_version, False
```
asfakianakis, line 49: Applied requested changes.

vortexasdk/client.py
```diff
@@ -195,6 +196,8 @@ def create_client() -> VortexaClient:
         raise KeyError(
             "VORTEXA_API_KEY environment variable is required to use the VortexaSDK"
         )
+    _ = warn_user_if_sdk_version_outdated()
+
```
asfakianakis, line 200: Applying requested changes.

vortexasdk/client.py
```diff
@@ -195,6 +196,8 @@ def create_client() -> VortexaClient:
         raise KeyError(
             "VORTEXA_API_KEY environment variable is required to use the VortexaSDK"
         )
+    _ = warn_user_if_sdk_version_outdated()
```
KitBurgess, line 199:
```suggestion
    warn_user_if_sdk_version_outdated()
```

KitBurgess: We don't need to assign anything here

vortexasdk/client.py
KitBurgess: Nice

PR #199, opened 2020-05-27
feat: add Notebook links to docs

### LINKED ISSUE #172: Add links to jupyter notebooks examples directory in the docs
It'd be good to have a link to the examples directory containing the more detailed jupyter notebooks somewhere here:

image

https://github.com/VorTECHsa/python-sdk/tree/master/docs/examples

asfakianakis:
Add links for notebooks in the Docs. This should close #172.


asfakianakis:
> When running `./docs/scripts/serve_docs.sh` locally, It doesn't look like this page can be seen. You'll need to edit this file https://github.com/VorTECHsa/python-sdk/blob/master/pydocmd.yml

Yes, was getting this when building the docs:

```
INFO    -  The following pages exist in the docs directory, but are not included in the "nav" configuration:

  - config/config.md
  - examples/7_Notebooks.md 
```

KitBurgess:
image


I don't think this is how it's intended to look?

KitBurgess:
Given that there's no python code needed for this markdown notebook page, we could create a markdown file directly. We could create the page in the same way we've done so for the [About Endpoints](https://github.com/VorTECHsa/python-sdk/blob/master/docs/endpoints/about-endpoints.md) page

asfakianakis:
> Given that there's no python code needed for this markdown notebook page, it'll probably be easiest to create a markdown file directly. We could create the page in the same way we've done so for the [About Endpoints](https://github.com/VorTECHsa/python-sdk/blob/master/docs/endpoints/about-endpoints.md) page, which ends up looking like this https://vortechsa.github.io/python-sdk/endpoints/about-endpoints/

Yes that's a goo idea. And will keep the same text as above, so just exposing the links, right?

KitBurgess:
I think we should include a link to just the https://github.com/VorTECHsa/python-sdk/tree/master/docs/examples page, else this list will definitely get out of date

KitBurgess:
Picked this back up - it looks like this: 
image

PR #209, opened 2020-06-17
docs: Add loading external data example

KitBurgess:


docs/examples/7_custom_excel_charterer_ingestion.py
```diff
@@ -0,0 +1,53 @@
...
+
+def convert_to_corporation_ids(corporation_name: str) -> List[ID]:
+    """
+    Returns a list of all corporation IDs with the name `corporation_name`.
+    This function returns a list, instead of a single ID. This is because some charterers may share the same name.
+
+    :param corporation_name: The name of the corporation we'd like to look up
+    :return: A list of IDs of corporations with a matching name
+    """
+    return [c.id for c in Corporations().search(corporation_name).to_list() if c.name == corporation_name]
```
gssantos, line 35: Instead of an exact string match, it might be better to do "c.name.upper() == corporation_name.upper()" or perhaps "corporation_name.upper() in event c.name.upper()"

PR #213, opened 2020-06-29
Allow users to interactively enter API key

KitBurgess:
Allow users to enter an API key like this: 

image

The API key is stored for the duration of the session, be this a python console, jupyter notebook, or a python script. Users can still use env vars for api keys.


vortexasdk/client.py
```diff
@@ -210,7 +209,7 @@ def set_client(client) -> None:
     )
 
 
-def warn_user_if_sdk_version_outdated() -> None:
+def _warn_user_if_sdk_version_outdated() -> None:
```
KitBurgess, line 212: made this a private function while we're here

KitBurgess:
> That's really nice feature, especially for non-tech people / Windows users (which is not as straight forward to set env variables).
> 
> One question: How do we know that the API key is stored for the whole session?

Good Q - the reason is this method here: https://github.com/VorTECHsa/python-sdk/blob/master/vortexasdk/client.py#L180

We store the client as a global variable, shared among all endpoints (Products, Geographies, VesselMovements etc), which is created once then reused.

PR #214, opened 2020-06-29
feat: Add exact match filter to geogs, vessels, corporations

### LINKED ISSUE #166: Better interface for retrieving IDs
```
crude = [p.id for p in Products().search('crude').to_list() if p.name=='Crude']
assert len(crude) == 1
```

This is the painful legwork we have to do currently to retrieve a Crude ID

KitBurgess:
This PR closes #166 , meaning you can now do:

`p = Geographies().search("China", exact_term_match=True)`

rather than previously:

`g = [g for g in Geographies().search("China") if g["name"] == "China"]`


tests/test_utils.py
```diff
@@ -5,27 +5,49 @@
...
     convert_values_to_list,
     get_latest_sdk_version,
+    filter_exact_match,
 )
 
 
 class TestUtils(TestCase):
     def test_single_item_list(self):
-        assert convert_to_list("a") == ["a"]
+        self.assertEqual(convert_to_list("a"), ["a"])
```
KitBurgess, line 14: just cleaning up some tests while we're here

tests/test_utils.py
```diff
@@ -5,27 +5,49 @@
...
+        self.assertCountEqual(convert_to_list(["a", "b"]), ["a", "b"])
 
     def test_convert_values_to_list(self):
-        d = {1: None, 2: "b", 3: ["c"]}
+        actual = convert_values_to_list({1: None, 2: "b", 3: ["c"]})
 
         expected = {1: [], 2: ["b"], 3: ["c"]}
 
-        assert convert_values_to_list(d) == expected
+        assert actual == expected
```
KitBurgess, line 24: just cleaning up some tests while we're here

vortexasdk/endpoints/cargo_movements.py
```diff
@@ -219,7 +219,7 @@ def search(
             ),
         }
 
-        params = {
+        api_params = {
```
KitBurgess, line 222: rename `params` to `api_params` for clarity

vortexasdk/endpoints/products.py
```diff
@@ -68,19 +68,16 @@ def search(
...
             "allowTopLevelProducts": True,
         }
 
-        search_result = super().search(**search_params)
-
-        if exact_term_match:
-            return ProductResult(filter_exact_match(term, search_result))
-        else:
-            return ProductResult(search_result)
+        return ProductResult(
```
KitBurgess, line 78: this logic now exists in the `Search` class, so that Vessels, Corporations, Geographies can use it without duplicating the code.

tests/test_utils.py
```diff
@@ -5,27 +5,49 @@
...
+
+    def test_filter_exact_match_single_name(self):
+        search_result = [{"name": "China"}, {"name": "South China"}]
+
+        actual = filter_exact_match("China", search_result)
+
+        expected = [{"name": "China"}]
+
+        self.assertEqual(actual, expected)
+
```
gssantos, line 41: We also need a test for filter_exact_match correctly returning an empty set (e.g. searching for "Chi")

PR #215, opened 2020-06-29
ci: Kill process on old docs if it exists

KitBurgess:
Boring piece of extra automation that makes it now a one-click process to update the documentation again.


docs/scripts/deploy_docs.sh
```diff
@@ -9,7 +9,15 @@ python docs/autogen.py
...
 
+
 #   1. serve the docs locally
+
+# Kill the any process running on port 8000 if it exists, this is the port that pydocmd serves on.
+process_on_our_port=$(sudo lsof -t -i:8000)
+if [[ $process_on_our_port ]]; then
+  sudo kill $process_on_our_port
+fi
+
```
gssantos, line 20: This is fine but you can also do it in one line:

process_on_our_port=$(sudo lsof -t -i:8000) && sudo kill $process_on_our_port

KitBurgess: Nice - I'll add that

PR #216, opened 2020-06-29
Password docs

KitBurgess:
Add documentation about the new interactive API key prompt, and update the check_setup script to reflect the changes.

### DIFF
--- a/vortexasdk/__init__.py
+++ b/vortexasdk/__init__.py
@@ -13,3 +13,6 @@ from vortexasdk.endpoints import (
 
 # noinspection PyUnresolvedReferences
 from vortexasdk.version import __version__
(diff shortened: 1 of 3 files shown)


vortexasdk/__init__.py
```diff
@@ -13,3 +13,6 @@
 
 # noinspection PyUnresolvedReferences
 from vortexasdk.version import __version__
+
+# noinspection PyUnresolvedReferences
+from vortexasdk.check_setup import run_all_checks
```
KitBurgess, line 18: while we're here, i've added this to the init file.

This allows you to write:

```
import vortexasdk as v
v.run_all_checks()
```

PR #219, opened 2020-07-14
feat: Floating storage beginner notebook

Natday:


KitBurgess:
Shall we also include the output of the notebook cells here? I think some users won't run the notebook, they'll just read through and view the input / output. For these people it'll be useful for them to see, plus it makes it a more complete documentation

Issue #195, opened 2020-05-12
Show correct exceptions to clients


KitBurgess:
We're raising an internal KeyError here (implying the SDK is at fault), when we should actually propogate the 400 error code (faulty input) to the user.

![image](https://user-images.githubusercontent.com/33626692/81671413-e0e2d000-9440-11ea-92f9-9289cc04e46e.png)

PR #225, opened 2020-07-29
[FEAT] Raise value error on server response error

laurence-ashdown:
Raises a `ValueError` to return to the user, with the response code, reason, and message if it exists

resolves: https://github.com/VorTECHsa/python-sdk/issues/195

### DIFF
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -38,7 +38,7 @@ class TestClient(TestCase):
         sample_valid_key = "123e4567-e89b-12d3-a456-426614174000"
         verify_api_key_format(sample_valid_key)
 
-    def test__raised_error_for_response_error(self):
+    def test_raised_error_for_response_error(self):
         badResponse = Response()
         badResponse.code = "bad request"
         badResponse.reason = "bad request"
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -1,6 +1,12 @@
 from unittest import TestCase
 
-from vortexasdk.client import VortexaClient, verify_api_key_format
+from vortexasdk.client import (
+    VortexaClient,
+    verify_api_key_format,
+    _handle_response,
+)
+
+from requests.models import Response
 
 
 class TestClient(TestCase):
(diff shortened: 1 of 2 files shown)


tests/test_client.py
```diff
@@ -31,3 +37,10 @@ def test_raised_error_for_invalid_uuid(self):
     def test_accepts_valid_uuid(self):
         sample_valid_key = "123e4567-e89b-12d3-a456-426614174000"
         verify_api_key_format(sample_valid_key)
+
+    def test__raised_error_for_response_error(self):
```
KitBurgess, line 41: What does the double underscore here signify?

vortexasdk/client.py
```diff
@@ -154,15 +154,18 @@ def _handle_response(response: Response, payload: Dict = None) -> Dict:
         logger.error(response.reason)
         logger.error(response.status_code)
         logger.error(response)
+        error = "[" + str(response.status_code) + " " + response.reason + "] "
```
KitBurgess, line 157: `f"{variable name} with surrounding sentence written here"` would be a nice way to do this

vortexasdk/client.py
```diff
@@ -154,15 +154,18 @@ def _handle_response(response: Response, payload: Dict = None) -> Dict:
...
+        message = ""
 
         # noinspection PyBroadException
         try:
             logger.error(response.json())
+            message = response.json()["message"]
         except Exception:
             pass
```
KitBurgess, line 165: let's set `message=""` here, in the case that message can't be extracted from the response, rather than define it line 158 and then mutate it

vortexasdk/client.py
```diff
@@ -154,15 +154,18 @@ def _handle_response(response: Response, payload: Dict = None) -> Dict:
...
 
         logger.error(f"payload: {payload}")
-        json = {}
+        raise ValueError(error + message)
```
KitBurgess, line 168: We can define the error + message string inline here, closer to the code call site, rather than further up in line 157

PR #227, opened 2020-08-04
feat: Add diversions endpoint

KitBurgess:


tests/endpoints/test_vessel_diversions_real.py
```diff
@@ -0,0 +1,20 @@
...
+class TestVesselDiversions(TestCaseUsingRealAPI):
+    def test_search(self):
+
+        crude_id = "6f11b0724c9a4e85ffa7f1445bc768f054af755a090118dcf99f14745c261653"
+        rotterdam = Geographies().search("Rotterdam [NL]", exact_term_match=True).to_list()[0].id
+
+        diversions = VesselDiversions().search(
+            filter_time_min=datetime.now() - timedelta(days=1),
+            # filter_locations=rotterdam,
+            # filter_products=crude_id,
```
SeanBarry, line 16: Just a reminder to not use any product filters as some user may not have permissions for these

PR #230, opened 2020-08-05
feat: Format timestamps as timestamps

### LINKED ISSUE #49: Dataframe timestamp columns in datetime format
linked to #45 
would be nice to have and would save additional manipulation in the notebook

KitBurgess:
Closes #49

### DIFF
--- a/vortexasdk/result_conversions.py
+++ b/vortexasdk/result_conversions.py
@@ -16,9 +16,10 @@ def create_list(list_of_dicts, output_class: FromDictMixin) -> List:
 
 def format_datatypes(df: pd.DataFrame) -> pd.DataFrame:
     """Format the relevant columns with sensible datatypes"""
-    timestamp_cols = [c for c in df.columns if "timestamp" in c]
+    timestamp_cols = [col for col in df.columns if "timestamp" in col]
 
-    df[timestamp_cols] = df[timestamp_cols].apply(pd.to_datetime)
+    for col in timestamp_cols:
+        df[col] = pd.to_datetime(df[col])
 
     return df
 
--- a/vortexasdk/result_conversions.py
+++ b/vortexasdk/result_conversions.py
@@ -14,6 +14,15 @@ def create_list(list_of_dicts, output_class: FromDictMixin) -> List:
     return [output_class.from_dict(d) for d in list_of_dicts]
 
 
+def format_datatypes(df: pd.DataFrame) -> pd.DataFrame:
+    """Format the relevant columns with sensible datatypes"""
+    timestamp_cols = [c for c in df.columns if "timestamp" in c]
+
+    df[timestamp_cols] = df[timestamp_cols].apply(pd.to_datetime)
+
+    return df
+
+
 def create_dataframe(
     columns: Union[None, List[str]],
     default_columns: List[str],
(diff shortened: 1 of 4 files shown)


vortexasdk/result_conversions.py
```diff
@@ -14,6 +14,15 @@ def create_list(list_of_dicts, output_class: FromDictMixin) -> List:
     return [output_class.from_dict(d) for d in list_of_dicts]
 
 
+def format_datatypes(df: pd.DataFrame) -> pd.DataFrame:
+    """Format the relevant columns with sensible datatypes"""
+    timestamp_cols = [c for c in df.columns if "timestamp" in c]
+
+    df[timestamp_cols] = df[timestamp_cols].apply(pd.to_datetime)
```
gssantos, line 21: Pandas does not allow you to set multi-column slices like this (see example below). You need to do:

for timestamp_col in timestamp_cols:
    df[timestamp_col] = pd.to_datetime(df[timestamp_col])

example:
a = pd.DataFrame({'a': [0, 1], 'b': [2, 3], 'c': [4, 5]})
a[['a', 'b']] = a[['a', 'b']] + 1
print(a)
   a  b  c
0  1  3  4
1  2  4  5

KitBurgess: Using the apply method does allow this - it's a different implementation to `a[['a', 'b']] = a[['a', 'b']] + 1`.

I've added a test though to confirm this works.

vortexasdk/result_conversions.py
```diff
@@ -30,8 +39,10 @@ def create_dataframe(
...
-        return pd.DataFrame(data=data, columns=default_columns)
+        df = pd.DataFrame(data=data, columns=default_columns)
     elif columns == "all":
-        return pd.DataFrame(data=data)
+        df = pd.DataFrame(data=data)
     else:
-        return pd.DataFrame(data=data, columns=columns)
+        df = pd.DataFrame(data=data, columns=columns)
+
+    return format_datatypes(df)
```
gssantos, line 48: format_datatypes() currently returns nothing, so create_daframe will return None

vortexasdk/result_conversions.py
```diff
@@ -30,8 +39,10 @@ def create_dataframe(
...
 
     if columns is None:
-        return pd.DataFrame(data=data, columns=default_columns)
+        df = pd.DataFrame(data=data, columns=default_columns)
     elif columns == "all":
-        return pd.DataFrame(data=data)
+        df = pd.DataFrame(data=data)
     else:
-        return pd.DataFrame(data=data, columns=columns)
+        df = pd.DataFrame(data=data, columns=columns)
```
cvonsteg, line 46: Could simplify this slightly

```python
columns = columns or default_columns # handles None 'if' clause
if columns == 'all':
    df = pd.DataFrame(data=data)
else:
    df = pd.DataFrame(data=data, columns=columns)
```

vortexasdk/result_conversions.py
cvonsteg: Note on this - Although it incurs the overhead of a loop, a more efficient implementation is probably:

```python
for col in timetstamp_cols:
    pd.to_datetime(df[c])
```

As you now no longer have the apply overhead and applies on dataframes generally aren't all that efficient (checked on a local sample df, this is about 5x faster than the df apply)

vortexasdk/result_conversions.py
KitBurgess: This is definitely fewer lines of code - but I think it's slightly more difficult to read for the user: we mutate the columns variable and then handle it in the if statement - rather than explicitly dealing with the three posibilities for the columns variable..

cvonsteg: You call!

PR #232, opened 2020-08-05
docs: Add exact term match to examples

KitBurgess:


docs/examples/3_crude_from_saudi_arabia_to_india.py
```diff
@@ -15,31 +15,32 @@
...
     all_geogs_with_india_in_the_name = Geographies().search("india").to_list()
 
+    # If running interactively, you may want to print all the names here to inspect them for yourself
+    for g in all_geogs_with_india_in_the_name:
+        print(g.name)
+
     # We're only interested in the country India here
     india = [
         g.id for g in all_geogs_with_india_in_the_name if g.name == "India"
     ]
```
asfakianakis, line 37: Should we use the layer option here? I.e. `if g.layer == "country"`

KitBurgess: We could do - I wanted to highlight here that that this is the second way to filter on exactly the name - by doing it manually yourself. I didn't mean the 'we want the country India here' comment to focus to much that the plolygon is a country layer.

asfakianakis: Fair enough

PR #234, opened 2020-08-05
ci: Test SDK against multiple pandas versions

KitBurgess:
We introduce tox here, allowing us to test the SDK against pandas < 1.0 and pandas >= 1.0. This is useful for us, as we've got SDK users spanning both 0.X and 1.X pandas versions.

Tox is used to automate the running of python tests against multiple package versions, dependencies, and multiple python versions. It’s normally used to test libraries.


setup.py
```diff
@@ -25,7 +25,7 @@
     ],
     python_requires=">=3.7",
     install_requires=[
-        "pandas==0.25.2",
+        "pandas>=0.25.2",
```
cvonsteg, line 28: Are there any null-type handling parts where we need to be aware of the `NA` type change in Pandas 1?

KitBurgess: This PR runs all SDK tests against both specified versions of pandas (0.25.2 and 1.0.0), so there's nothing we're aware of yet.

cvonsteg: tox ftw

PR #241, opened 2020-08-27
fix: set all_test_pass to False if API KEY not found

cvonsteg:
`check_setup.py` script does not set `all_test_pass` flag to False if the API key is not found.

Solution: Set to false if not found.


cvonsteg:
Jumped the gun on this one 😢

PR #246, opened 2020-10-16
fix: Edit failing test

KitBurgess:
The test that was removed was failing. This isn't the unit test's fault, but merely the fact that these `TestCaseUsingRealAPI` unit tests rely on an external data source - the Vortexa API.

These conversion tests should instead use a mock api, like other tests extending from the `TestCaseUsingMockAPI` in the SDK.

In the meantime, we edit the faulty test to ensure development can continue.


SeanBarry (approved):

SeanBarry:
LGTM

asfakianakis (approved):

PR #247, opened 2020-10-29
Add support for EIA forecasts endpoint

dufia:
Adds the EIA forecast endpoint.


vortexasdk/api/eia_forecast.py
```diff
@@ -0,0 +1,17 @@
+from dataclasses import dataclass
+from typing import List
+
+from vortexasdk.api.serdes import FromDictMixin
+from vortexasdk.api.shared_types import IDName
+
+
+@dataclass(frozen=True)
+class EIAForecast(IDName, FromDictMixin):
```
KitBurgess, line 9: This `EIAForecast` class currently inherits from IDName - which should only be the case if each EIAForecast record returns by the API has both an `id` attribute and a `name` attribute. @dufia This is likely the cause of the failing tests

dufia: solved 👍 thanks

vortexasdk/endpoints/eia_forecasts.py
```diff
@@ -0,0 +1,80 @@
...
+
+    """
+
+    def __init__(self):
+        Search.__init__(self, EIA_FORECASTS_RESOURCE)
+
+    def search(
+        self,
+        preset: str = None,
+        filter_time_min: str = False,
```
KitBurgess, line 28: The type here should be datetime, rather than string. We'll also want sensible default datetimes rather than the bool `False`

vortexasdk/endpoints/eia_forecasts.py
```diff
@@ -0,0 +1,80 @@
...
+class EIAForecasts(Search):
+    """EIA forecasts Endpoint, use this to search through Vortexa's EIA Forecasts data.
+
+    The data includes:
+
+    - Vortexa's data science based forecast of the EIA number to be published on the week
+    - Actual EIA import/export numbers as published by the EIA Weekly Supply Estimates report
+    - EIA stocks (kbl)
+    - Cover (days of Supply for the whole of the US, as published by the EIA Weekly Supply Estimates report)
+    - Refinery runs (refiner “Percent Operable Utilization” as published by the EIA Weekly Supply Estimates report)
```
KitBurgess, line 18: Nice - should we make this even clearer by specifying the column name first, then the accompanying documentation? something like:

```runs: Refinery runs (refiner...```

this means it'll be easier for clients to lookup the functionality associated with a col name

File: vortexasdk/endpoints/eia_forecasts_result.py [hunk 4]
```diff
@@ -0,0 +1,41 @@
...
+        """Represent EIAForecast data as a list."""
+        # noinspection PyTypeChecker
+        return create_list(super().to_list(), EIAForecast)
+
+    def to_df(self, columns=None) -> pd.DataFrame:
+        """
+        Represent EIA forecasts as a `pd.DataFrame`.
+
+        # Arguments
+            columns: The EIA forecasts features we want in the dataframe. Enter `columns='all'` to include all features.
```
KitBurgess, line 26:
```suggestion
            columns: The EIA forecasts columns we want in the dataframe. Enter `columns='all'` to include all columns.
```

KitBurgess: In the world of datascience, the term features normally refers to the input features of a model. In this sense of the word, we're better off using 'column' rather than features, as these aren't input features.

vortexasdk/endpoints/eia_forecasts.py
```diff
@@ -0,0 +1,80 @@
...
+        Search.__init__(self, EIA_FORECASTS_RESOURCE)
+
+    def search(
+        self,
+        preset: str = None,
+        filter_time_min: str = False,
+        filter_time_max: str = False,
+    ) -> EIAForecastResult:
+        """
+        Find EIA forecasts for a given date preset and date range.
```
KitBurgess, line 32: Do we mean date preset here? It's more of a geography & product preset right?

vortexasdk/endpoints/eia_forecasts.py
```diff
@@ -0,0 +1,80 @@
...
+        ```
+
+        returns
+
+        | date                     | forecast_fri     | value | stocks | cover | runs |
+        | ------------------------ | ---------------- | ----- | ------ | ----- | ---- |
+        | 2020-01-31T00:00:00.000Z | 454.96048964485  | 323   | 9541   | 26.5  | 65.9 |
+        | 2020-01-24T00:00:00.000Z | 545.453497230504 | 579   | 10461  | 25.9  | 61.5 |
+        | 2020-01-17T00:00:00.000Z | 510.289752707662 | 549   | 10325  | 25.2  | 64.7 |
+        | 2020-01-10T00:00:00.000Z | 469.841470826967 |       |        |       |      |
```
KitBurgess, line 70: Should we give an accompanying note as to why some of the values are null? Clients will definitely ask this Q - we may as well answer it here once.

vortexasdk/endpoints/eia_forecasts.py
```diff
@@ -0,0 +1,80 @@
...
+        self,
+        preset: str = None,
+        filter_time_min: str = False,
+        filter_time_max: str = False,
+    ) -> EIAForecastResult:
+        """
+        Find EIA forecasts for a given date preset and date range.
+
+        # Arguments
+            preset: the EIA forecasts preset to be returned. Preset can be: 'padd1-gasoline-imports',
```
KitBurgess, line 35: Should we explain what a preset is?

KitBurgess (changes requested):
We'll also want to update the about-endpoints page that lists the available endpoints.

just notices the cargo time series endpoint is missing - while we're here can we add that?

 http://localhost:8000/endpoints/about-endpoints/

vortexasdk/endpoints/eia_forecasts.py
dufia: Fixed, it should've been: "Find EIA forecasts for a given preset and date range."

KitBurgess (approved):

PR #249, opened 2020-11-17
docs: Run through the deployment process with @dufia

KitBurgess:


KitBurgess:
thankfully this is irrelevant now as the process is automated

PR #262, opened 2021-01-07
Introduce dependabot

KitBurgess:


vortexasdk/version.py
```diff
@@ -1 +1 @@
-__version__ = "0.28.0"
+__version__ = "0.28.1"
```
Christos-Hadjinikolis, line 1: For my benefit: this is irrelevant to dependabot right?

Christos-Hadjinikolis (approved):
LGTM

vortexasdk/version.py
KitBurgess: Yes - this is required to deploy a new version of the SDK & documentation

PR #268, opened 2021-01-07
chore: We do not need the six dependency

KitBurgess:
dependabot prompted me to look into how we're using the `six` library (which provides backwards compatability between python2 and 3). We don't need to cater for python 2, so can remove it.


Christos-Hadjinikolis (approved):
LGTM

PR #299, opened 2021-06-07
fix: tests

yannickrocks:
Resolves https://github.com/VorTECHsa/app-issues/issues/1060


tests/endpoints/test_cargo_movements_real.py
```diff
@@ -187,7 +187,7 @@ def test_search_single_filter_owner_name(self):
             .to_df()
             .head(2)
         )
-        assert len(df) == 2
+        assert len(df) == 1
```
ChristianBorresen, line 190: So we now have only one CM in that time frame for DHT, and previously we had two?

ChristianBorresen (changes requested):

tests/endpoints/test_cargo_movements_real.py
yannickrocks: I believe so. I think the reason the data change might be when we changed the indices.

ChristianBorresen: Cool. Terrible test though.

KitBurgess: Terrible terrible

vortexasdk/version.py
```diff
@@ -1 +1 @@
-__version__ = "0.28.8"
+__version__ = "0.28.9"
```
RaduG, line 1: Does this actually warrant a version bump? Tests are not really part of the package clients get.

ChristianBorresen (approved):

PR #313, opened 2021-08-31
Support array of statuses

Selerski:


vortexasdk/endpoints/vessel_movements.py
```diff
@@ -185,7 +185,7 @@ def search(
             "filter_products": convert_to_list(filter_products),
             "filter_vessels": convert_to_list(filter_vessels),
             "filter_vessel_classes": convert_to_list(filter_vessel_classes),
-            "filter_vessel_status": filter_vessel_status,
+            "filter_vessel_status": convert_to_list(filter_vessel_status),
```
Selerski, line 188: this param only works when it's an array, so this converts a string to an array

JRGranell (approved):

PR #322, opened 2021-10-14
Unpin dependencies

RaduG:
The package dependencies of this library are pinned, which is fine in a majority of cases if used exclusively in a notebook environment, but suboptimal if part of a production process. The direct effect of this limitation is that, for example, one is not able to use newer versions of libraries that require `requests`.

Of course, unpinning should be done in a controlled way, ensuring that the code still works in any permutation of dependency versions. Therefore, I propose the following changes:

### 1. Unpin dependencies
From:
```
"flatten-dict==0.2.0",
 "jsons==1.0.0",
 "jupyter==1.0.0",
 "matplotlib==3.3.4",
 "pandas>=0.25.2",
 "requests==2.22.0",
 "statsmodels==0.12.2",
 "tqdm==4.38.0",
```

To:
```
"flatten-dict~=0.2.0",
 "jsons~=1.0.0",
 "jupyter~=1.0.0",
 "matplotlib~=3.3.4",
 "pandas>=0.25.2,<=1.3.0",
 "requests>=2.22.0,<=2.26.0",
 "statsmodels~=0.12.2",
 "tqdm~=4.38.0",
```

Some dependencies are relaxed to `compatible`, some are given a range. In any case, `pandas` and `requests` are the main pain-points.

### 2.  Test more comprehensively with tox
In the current setup, we only test two scenarios - pandas 0.25 and pandas 1.0 (so we run the tests twice). Given the range of versions the requirements have been relaxed to, we should instead test all combinations of minor versions in the range of supported versions for `pandas` and `requests`. This means now we have to run the test suite 25 times.

The result of this is that, even with the addition of the `-o` flag to tox, which executes tests environments in parallel (albeit with only 2 instances since our provisioned CI executor has 2cpus), testing takes over 2 hours.

What can we do to address this? Well, I think we have a couple of options:
1. Test fewer scenarios (maybe fix the version of `requests` for all the `pandas` variations and vice-versa, bringing the number down to 10 suites).
2. Increase the size of the executor. Assuming almost linear scaling, we should be able to bring this down to 30 mins (from the original ~15 before this PR) with an 8-cpus instance.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -129,7 +129,7 @@ jobs:
           command: |
             . .venv/bin/activate
             pip install tox
-            tox  --parallel --verbose
+            tox  -o --verbose
 
   check-release-version:
     working_directory: ~/project
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -118,7 +118,7 @@ jobs:
     working_directory: ~/project
     docker:
       - image: circleci/python:3.7
-    resource_class: 2xlarge
+    resource_class: xlarge
     steps:
       - checkout
       - restore_cache: &restore_cache
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -118,6 +118,7 @@ jobs:
     working_directory: ~/project
     docker:
       - image: circleci/python:3.7
+    resource_class: 2xlarge
     steps:
       - checkout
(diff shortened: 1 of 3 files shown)


SeanBarry:
LGTM

PR #326, opened 2021-11-16
Pricing docs

Selerski:
see https://github.com/VorTECHsa/app-issues/issues/1332


.circleci/config.yml
```diff
@@ -53,7 +53,7 @@ jobs:
           command: |
             . .venv/bin/activate
             # This should be done in tox, but i can't work out how to run doctests in tox hence we're also running it here for now.
-            pytest -v --doctest-modules
+            pytest -v --doctest-modules --ignore=tests
```
Selerski, line 56: ensures we don't run API tests in doc tests

vortexasdk/endpoints/endpoints.py
```diff
@@ -31,5 +31,8 @@
 FLEET_UTILISATION_TIMESERIES = '/vessel-journey/timeseries/utilisation'
 FLEET_UTILISATION_TIMESERIES_AVG_DISTANCE = '/vessel-journey/timeseries/avg-distance'
 
-API_FQDN = "https://api.vortexa.com"
+FREIGHT_PRICING_SEARCH = '/freight-pricing/rates'
+FREIGHT_PRICING_TIMESERIES = '/freight-pricing/timeseries'
+
+API_FQDN = "https://api-feat-rates-doc.public.kube.develop.vortexa.com"
```
Selerski, line 37: restore before merge

vortexasdk/version.py
```diff
@@ -1 +1 @@
-__version__ = "0.32.1"
```
andreagerino, line 1: The patch number should be reset to 0, no? `0.33.0`

Selerski: done

PR #354, opened 2022-06-10
Raw json response from voyages

Selerski:


vortexasdk/api/voyages.py
```diff
@@ -59,17 +48,17 @@ class VoyagesVesselEntity():
 
     id: ID
     name: str
-    imo: int
+    imo: Optional[int]
```
Selerski, line 51: not all vessels have imo, year, propulsion, ice_class - this is contrary to the schema. Is this fine?

vortexasdk/api/shared_types.py
```diff
@@ -114,7 +115,7 @@ class Flag:
 
     tag: str
     flag: str
-    flag_country: str
+    flag_country: Optional[str]
```
Selerski, line 118: not all records have flag_country - is this ok?

vortexasdk/endpoints/voyages_search_enriched.py
```diff
@@ -294,3 +294,259 @@ def search(
         }
 
         return VoyagesSearchEnrichedResult(super().search(headers=self._HEADERS, **api_params))
+
+    def search_raw(
```
Selerski, line 298: `search_raw` doesn't send `text/csv` headers or `columns` API params to the server - the server returns only a JSON response.

vortexasdk/endpoints/voyages_search_enriched_result.py
```diff
@@ -36,3 +36,22 @@ def to_df(self) -> pd.DataFrame:
 
         # converts list to a datafrane
         return pd.DataFrame(data=super().to_list())
+
+
+class VoyagesSearchEnrichedRawResult(Result):
```
Selerski, line 41: `search` result - has `to_df()` (converts CSV to a df), doesn't have `to_list()`
`search_raw` result - has `to_list()` (converts a raw JSON response to a list), doesn't have `to_df()`

vortexasdk/api/voyages.py
andreagerino: Yeah, we should update the schema really...

vortexasdk/endpoints/voyages_search_enriched.py
```diff
@@ -293,4 +296,7 @@ def search(
             "vessel_risk_level_excluded": convert_to_list(vessel_risk_level_excluded),
         }
 
-        return VoyagesSearchEnrichedResult(super().search(headers=self._HEADERS, **api_params))
+        if columns is None:
+            return VoyagesSearchEnrichedListResult(super().search(**api_params))
+        else:
+            return VoyagesSearchEnrichedFlattenedResult(super().search(headers=self._HEADERS, **api_params))
```
andreagerino, line 302: I would rename `HEADERS` to `CSV_HEADERS` so there is no doubt about why we pass them only in one case.

andreagerino (changes requested):

vortexasdk/api/voyages.py
```diff
@@ -59,30 +39,98 @@ class VoyagesVesselEntity():
...
 
 
     """
+    # shared event properties
+    event_id: str
+    start_timestamp: Optional[ISODate]
+    end_timestamp: Optional[ISODate]
+    event_group: str
+    event_type: str
+    activity: str
```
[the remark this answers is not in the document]
Selerski: All `event_types` are defined enums.

Also [this](https://github.com/VorTECHsa/python-sdk/pull/354/files/c6e55eb3948792683e94444337119fe3f606bc80#diff-1e40b4fcbd34068fb00a8449f081b43d12bae206c7c6ca4674a430cf3f922d11R125) is misleading, I forgot to delete the comment. These are status event props

Now `activity` is undefined in cargo events, optional string in vessel event, and a string in status event

andreagerino (approved):

PR #356, opened 2022-07-06
Fix optional props

Selerski:


vortexasdk/api/voyages.py
```diff
@@ -160,5 +160,5 @@ class VoyageEnrichedItem(FromDictMixin):
     corporate_entities: List[Entity]
     odometer_start: Optional[int]
     odometer_end: Optional[int]
-    events: List[Optional[Union[VoyageStatusEvent, VoyageVesselEvent, VoyageCargoEvent]]]
```
Selerski, line 163: serialization for a type this complex doesn't work very well, so I cast this type as a dict.

To get properties of this object, one would have to look into the docs

PR #364, opened 2022-08-02
fix: search params - storage terminals endpoint

OllyDS:
They should be using the keyword `term` rather than `name`

Resolves: https://github.com/VorTECHsa/app-issues/issues/1871


vortexasdk/endpoints/storage_terminals.py
```diff
@@ -61,24 +61,24 @@ def load_all(self) -> StorageTerminalResult:
...
-        name: Union[str, List[str]] = None,
+        term: Union[str, List[str]] = None,
     ) -> StorageTerminalResult:
         """
-        Find all storage terminals matching given type.
+        Find all storage terminals matching given term.
 
         # Arguments
-            type: The type of storage terminal we're filtering on.
+            term: The term of storage terminal we're filtering on.
```
Selerski, line 70: Shouldn't ids be added to this description?

.all-contributorsrc
```diff
@@ -147,6 +147,15 @@
       "contributions": [
         "code"
       ]
+    },
```
Selerski, line 150: Nice ;)

Selerski (changes requested):

Selerski (approved):
if tests pass I'm happy to get this in

docs/examples/try_me_out/availability_search.ipynb
```diff
@@ -26,7 +26,18 @@
     },
     "tags": []
    },
-   "outputs": [],
+   "outputs": [
```
Selerski, line 29: Please clear these - this is going to put your file structure in the public domain :)

Selerski (changes requested):

docs/examples/try_me_out/availability_search.ipynb
```diff
@@ -94,7 +127,7 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3",
+   "display_name": "Python 3.8.11 ('venv': venv)",
```
Selerski, line 130: why does this need to change? same for below like with interpreter and hash

OllyDS: These must have been auto-generated as I didn't change these. Will see how to revert

Selerski (approved):

PR #367, opened 2022-08-22
[WIP] Drastic increase in performance when calling `to_list()`

SeanBarry:
Resolves https://github.com/VorTECHsa/app-issues/issues/1935

This PR involves a large number of file changes, but actually consists of a small number of core changes that cause the side effect of lots of small file changes.

The core changes are:
- Implement [Pydantic](https://pydantic-docs.helpmanual.io/). This is a very performant library for serializing/deserializing and schema validation. By converting all the dataclasses in the SDK to Pydantic models, we can easily deserialize results from the API in to Python classes which we can then call methods on. This deserialisation was the root cause of the performance issues on the `to_list()` method. [Here is the specific place in this PR](https://github.com/VorTECHsa/python-sdk/blob/to-list-perf-fix/vortexasdk/result_conversions.py#L14) where the root cause is fixed.
- A side effect of implementing Pydantic and getting free schema validation was the discovery that many of the schemas are outdated and incorrect. Many of these have now been updated, using TypeScript types as the source of truth.
- In order to prevent the SDK from throwing errors if the schema from the data from the API differs from the models defined in the SDK, all of the types in the Python SDK have been annotated as `Optional` with a default of `None`.

Other changes made to solve issues faced during this PR:
- Change CircleCI Cache key to use the checksum of `setup.py`. Previously, if a package version was changed, or a new package was added, this would not invalidate the dependency cache because the key was based on the branch name and initial build epoch
- Suppress missing type errors for `flatten_dict` and `tqdm`
- Add type stubs for some libraries

Results:
Example of `to_list()` before changes:
before

Example of `to_list()` after changes:
Screenshot 2022-08-30 at 22 05 16

All tests pass:
Screenshot 2022-08-31 at 14 19 02


setup.py
```diff
@@ -25,20 +25,20 @@
...
     install_requires=[
-        "flatten-dict==0.2.0",
-        "jsons==1.0.0",
+        "flatten-dict==0.4.2",
         "jupyter==1.0.0",
         "matplotlib==3.3.4",
         "pandas>=0.25.2",
-        "requests==2.27.1",
+        "requests==2.28.1",
         "statsmodels==0.12.2",
```
andreagerino, line 33: Consider updating `statsmodel` as this version doesn't compile under M1 (in another task)

tests/api/test_cargo_movement.py
```diff
@@ -24,111 +18,101 @@ class TestCargoMovement(TestCase):
...
+            assert self.cm == deserialized
 
     def test_convert_to_flat_dict(self):
-        flat = convert_cargo_movement_to_flat_dict(serialize_to_dict(self.cm))
+        flat = convert_cargo_movement_to_flat_dict(self.cm.dict())
 
         expected = {
             "cargo_movement_id": "00886b05a0747522b67322f50123ee60e61e219fc9a9c6011be1a1dade65f63e",
             "events.cargo_port_load_event.0.end_timestamp": "2019-10-20T16:41:49+0000",
+            "events.cargo_port_load_event.0.end_timestamp": datetime.datetime(
```
andreagerino, line 113: Do we need to create the `datetime` here?

tests/endpoints/test_asset_tanks.py
```diff
@@ -1,3 +1,4 @@
+from weakref import ref
```
andreagerino, line 1: Please check this import

tests/endpoints/test_cargo_movements_real.py
```diff
@@ -9,8 +9,7 @@
 
 class TestCargoMovementsReal(TestCaseUsingRealAPI):
     def test_default_search(self):
-        results = CargoMovements().search(filter_activity="loading_state")
-        print(len(results))
+        CargoMovements().search(filter_activity="loading_state")
```
andreagerino, line 12: There should be no param in the default search.

tests/endpoints/test_products.py
```diff
@@ -22,7 +22,20 @@ def test_serialize(self):
             ],
         }
 
-        Product.from_dict(crude_dict)
+        Product.parse_obj(crude_dict)
+
+    def test_serialize_with_missing_fields(self):
```
andreagerino, line 27: 🎉

vortexasdk/api/aggregation_breakdown_item.py
```diff
@@ -14,7 +12,7 @@ class AggregationBreakdownItem(FromDictMixin):
     and `value` and `count` correspond to numeric values of the returned record.
     """
 
-    id: str
-    count: int
-    value: float
-    label: str
+    id: Optional[str] = None
```
andreagerino, line 15: Why is this optional?

vortexasdk/api/attribute.py
```diff
@@ -1,15 +1,13 @@
...
+from pydantic import BaseModel
+
 
-from vortexasdk.api.serdes import FromDictMixin
 from vortexasdk.api.shared_types import Node
 
 
-@dataclass(frozen=True,)
-class Attribute(Node, FromDictMixin):
+class Attribute(Node, BaseModel):
```
andreagerino, line 8: `Node` inherits from `BaseModel` already, do we need to inherit from it here as well?

vortexasdk/api/cargo_movement.py
```diff
@@ -57,9 +148,27 @@ class CargoMovement(FromDictMixin):
...
-    events: List[CargoEvent]
-    parent_ids: List[ParentID]
+    quantity: Optional[int] = None
+    status: Optional[str] = None
+    vessels: Optional[List[VesselEntity]] = None
+    product: Optional[List[CargoMovementProductEntry]] = None
+    parent_ids: Optional[List[ParentID]] = None
+    events: Optional[
+        List[
+            Annotated[
```
andreagerino, line 158: What is the `Annotated` type?

vortexasdk/api/freight_pricing.py
```diff
@@ -1,24 +1,30 @@
...
-    cost_precision: int
-    cost_unit: str
-    tce: Optional[float]
-    tce_precision: Optional[int]
-    tce_unit: Optional[str]
-    source: Optional[str]
-    route_prediction: Optional[str]
+    short_code: Optional[str] = None
+    rate: Optional[float] = None
+    record_date: Optional[datetime] = None
```
andreagerino, line 21: should this be `ISODate` or `str`?

vortexasdk/api/shared_types.py
```diff
@@ -1,90 +1,118 @@
...
+from pydantic import BaseModel
 from datetime import datetime
 from typing import List, Optional, Union
 
 from vortexasdk.api.id import ID
 
 IDsNames = Union[List[Union[ID, str]], str, ID]
 
-ISODate = str
+ISODate = datetime
```
andreagerino, line 10: `ISODate` is just an alias for `datetime`? Worth considering given that we use them both in our types

vortexasdk/api/vessel.py
```diff
@@ -41,31 +61,30 @@ class Vessel(Node, FromDictMixin):
...
-@dataclass(frozen=True)
-class VesselEntity(IDName):
+class VesselEntity(IDName, BaseModel):
     """
     A VesselEntity represents a vessel record used in CargoMovements and VesselMovements.
 
     [Vessel Entities Further Documentation](https://docs.vortexa.com/reference/intro-vessel-entities)
     """
 
     id: ID
```
andreagerino, line 77: Probably not needed due to inheriting from `IDName`

vortexasdk/api/vessel.py
```diff
@@ -76,7 +95,7 @@ class VesselEntity(IDName):
     end_timestamp: Optional[ISODate] = None
     fixture_id: Optional[str] = None
 
-    scrubber: Optional[Scrubber] = None
-    flag: Optional[Flag] = None
+    scrubber: Optional[List[Scrubber]] = Field(default_factory=list)
```
andreagerino, line 98: I know you are treating this as interchangeable with `None` but I would prefer to be consistent and use the same approach.

vortexasdk/api/voyages.py
```diff
@@ -144,21 +160,27 @@ class VoyageEnrichedItem(FromDictMixin):
...
+    latest_product_ids: Optional[List[ID]] = Field(default_factory=list)
+    tags: Optional[List[Tag]] = Field(default_factory=list)
+    tonne_miles: Optional[int] = None
+    corporate_entities:Optional[List[EntityWithSingleLayerAndTimespan]] = Field(
+        default_factory=list
+    )
+    odometer_start: Optional[int] = None
+    odometer_end: Optional[int] = None
+    events: Optional[List[
+        Optional[Union[VoyageVesselEvent, VoyageCargoEvent, VoyageStatusEvent]]
```
andreagerino, line 182: Can you please add a comment about the fact that the order of the classes matters?

vortexasdk/endpoints/asset_tanks.py
```diff
@@ -108,4 +108,13 @@ def search(
             "term": [str(e) for e in convert_to_list(term)],
         }
 
-        return AssetTankResult(super().search(**search_params))
+        response = super().search_with_client(
```
andreagerino, line 111: Can you please explain why we need `search_with_client`?

vortexasdk/endpoints/freight_pricing_result.py
```diff
@@ -97,11 +97,10 @@ def to_df(self, columns=None) -> pd.DataFrame:
...
 DEFAULT_COLUMNS = [
-    'short_code',
-    'rate'
-    'rate_unit',
-    'cost',
-    'cost_unit',
-    'tce',
-    'tce_unit'
+    "short_code",
+    "rate" "rate_unit",
```
andreagerino, line 101: This is a bug

vortexasdk/endpoints/reference_breakdown_result.py
```diff
@@ -7,31 +7,33 @@
...
+from pydantic import BaseModel
 import pandas as pd
 import functools
 import os
 from multiprocessing.pool import Pool
 
 logger = get_logger(__name__)
 
+
 def replace_keys(result):
```
andreagerino, line 19: Please annotate `result`

vortexasdk/endpoints/voyages_search_enriched_result.py
```diff
@@ -21,9 +22,11 @@ class VoyagesSearchEnrichedFlattenedResult(Result):
 
     def to_list(self) -> List[VoyageEnrichedItem]:
         # noinspection PyTypeChecker
-        raise Exception(f"to_list method is not supported for search results in the flattened format (i.e. when the `columns` API param is provided). Please use to_df() instead.")
+        raise Exception(
+            f"to_list method is not supported for search results in the flattened format (i.e. when the `columns` API param is provided). Please use to_df() instead."
+        )
 
-    def to_df(self) -> pd.DataFrame:
+    def to_df(self, columns: Optional[List[str]] = None) -> pd.DataFrame:
```
andreagerino, line 29: Different from usual?

vortexasdk/operations.py
```diff
@@ -67,9 +68,13 @@ def __init__(self, resource):
         """
         self._resource = resource
 
-    def search(
-        self, exact_term_match: bool = None, response_type: str = None, headers: dict = None, **api_params
-    ) -> List[dict]:
+    def search_with_client(
```
andreagerino, line 71: Can you please add a code comment explaining why we need this?

setup.py
SeanBarry: https://github.com/VorTECHsa/app-issues/issues/1969

tests/endpoints/test_cargo_movements_real.py
SeanBarry: This predates my PR, but it's now optional.

vortexasdk/api/attribute.py
SeanBarry: Fixed in all places

vortexasdk/api/cargo_movement.py
SeanBarry: The Annotated type adds context specific metadata to a type. In this case, Pydantic is reading the metadata to know which field in the Model to use a discriminator. See docs here: https://pydantic-docs.helpmanual.io/usage/types/#discriminated-unions-aka-tagged-unions

vortexasdk/api/vessel.py
SeanBarry: Fixed.

vortexasdk/api/vessel.py
SeanBarry: Fixed.

vortexasdk/api/voyages.py
SeanBarry: Done.

vortexasdk/endpoints/freight_pricing_result.py
SeanBarry: Fixed in my other PR, but also fixed here now.

vortexasdk/endpoints/reference_breakdown_result.py
SeanBarry: Fixed.

vortexasdk/endpoints/voyages_search_enriched_result.py
SeanBarry: The inferred type was specifically annotated here, but I've removed it for consistency.

tests/api/test_cargo_movement.py
SeanBarry: I've changed this back to a str.

vortexasdk/api/freight_pricing.py
SeanBarry: I've made this ISODate (alias for str)

vortexasdk/operations.py
SeanBarry: In each dataset there is a `search` method that all of the endpoint parameters are passed to by users. This search method then combines all the parameters and passes them to the `search` method inherited from a parent class `Search`. These methods have the same method but different signatures. There's no need for them to have the same name and signature, so I've fixed this typing issue by just renaming the parent `search` method to avoid a signature collision.

An alternative would be to annotate the search methods such that they can accept any kwargs. This `search_with_client` method is not used by consumers so this is not a breaking change.

vortexasdk/endpoints/asset_tanks.py
SeanBarry: I've explained this in another comment, and added a code comment.

vortexasdk/operations.py
SeanBarry: I've added a code comment too

andreagerino (approved):

PR #375, opened 2022-09-13
Change data structure for latest products

Selerski:
see https://github.com/VorTECHsa/app-issues/issues/1084

Tests:
- ran all CI tests 
- ran tests locally (against the API with the new data structure)

We will have to notify customers to upgrade their SDKs. Once they've done that, we can release other parts of this PR


Selerski:
Tests will be failing until we get this change merged into API

vortexasdk/api/voyages.py
```diff
@@ -189,4 +189,4 @@ class VoyageEnrichedItem(BaseModel):
             ]
         ]
     ] = None
-    latest_product_details: Optional[List[EntityWithSingleLayer]] = None
```
andreagerino, line 192: What will happen to current SDK users when we release this fix?

andreagerino (changes requested):
Can we please add information about how we tested this to the PR description?

andreagerino:
What is going to happen when a user of the old PythonSDK hits the new `api` endpoint with array of arrays?

andreagerino:
What is going to happen when a user of the new PythonSDK hits the old api endpoint with just an array?

Selerski:
@andreagerino 

> What is going to happen when a user of the old PythonSDK hits the new api endpoint with array of arrays?
It is a breaking change for them and they will get a Pydantic validation error
> What is going to happen when a user of the new PythonSDK hits the old api endpoint with just an array?
This will work, because I've made latest products a union array and array of arrays

I think the best way to address this is:
- first release the Python SDK update with the union of two types
- In the last 30 days, Shell, Litasco, Hafnia, Navig8 and ADNOC have used this endpoint - reach out to them and force the upgrade.
- Once they've upgraded, we can release the rest of the PRs 
- In the long run, we'll also want to remove the union and make it a list of lists

Screenshot 2022-09-30 at 12 22 06

vortexasdk/api/voyages.py
Selerski: see my comment in the main thread

PR #376, opened 2022-09-22
fix(types): corrected param types for search enriched

cojdev:
https://github.com/VorTECHsa/app-issues/issues/1986


Selerski:
@cojdev Some tests are still failing, could you please have a look before I review?

Selerski:
You've committed the metadata of your latest test run into GH, but this is not necessary. Can you please remove the changes to the python notebooks metadata?

You only need two file changes really - the version file and the typing. It would be nice to add tests to check if both params work as intended, could you please add them?

tests/endpoints/test_voyages_search_enriched.py
```diff
@@ -75,3 +75,33 @@ def test_search_from_description_raw(self):
         assert res[0].vessel is not None
         assert len(res[0].events) > 0
         assert res[0].events[0].event_group is not None
+
+    def test_has_charterer_param(self):
```
Selerski, line 79: nice, thanks for adding those

vortexasdk/endpoints/voyages_search_enriched.py
```diff
@@ -5,6 +5,7 @@
 """
 from datetime import datetime
 from typing import Any, Dict, List, Union
+from typing_extensions import Literal
```
Selerski, line 8: This change made me realise we're being too lenient with our API param type definitions. For example, `vessel_risk_level: Union[str, List[str]] = None,` ->  we say this may be any random string, but this is not quite right

Could you please create a follow up ticket to:
- check all endpoints for params which are not strictly typed (mainly strings, but there may be others)
- introduce a similar Literal type wherever appropriate

You can add this ticket to one of the debt work packages, someone will pick it up at a later time

Selerski (approved):

PR #387, opened 2022-10-24
feat: added filter_layer param to geo ref endpoint

jakubbuskiewicz:
https://vortexa.atlassian.net/browse/ART-12


vortexasdk/endpoints/geographies.py
```diff
@@ -39,6 +40,8 @@ def search(
                 ['China', 'South China', 'China Energy Services Ningbo'...] etc. When `exact_term_match=True`,
                 the SDK will only yield the geography named `China`.
 
+            filter_layer: Must be one of geographical type ['scrapyard', 'anchoring', 'berth', 'terminal', 'port', 'country', 'shipping_region', 'region', 'trading_block', 'trading_region', 'trading_subregion', 'sts_zone', 'waypoint', 'storage', 'storage_terminal', 'fragment', 'root', 'legacy_sr'].
```
ChristianBorresen, line 43: I think we have to remove scrapyard, anchoring, berth, fragment, and possibly legacy_sr. Can you test if those even works?

ChristianBorresen (changes requested):

vortexasdk/endpoints/geographies.py
ChristianBorresen: I believe those are restricted

PR #425, opened 2023-03-20
feat: Remove non-core dependencies to extras section

joloppo:
## CHANGELOG

- Moved the packages jupyter, matplotlib and statsmodels to the extras section of the setup. 

These are not needed for the core sdk, and are only used for the demo notebooks. They bloat package install by >50MB.

## TESTS

- Testing currently fails on one test, probably unrelated. Please advise.
image

## COMMENTS

- Please advise on whether this makes sense and release process. 
- Related ticket: https://vortexa.atlassian.net/browse/RND-3595
- Thread: https://vortexa.slack.com/archives/CE4C35PN3/p1694595674073659


setup.py
```diff
@@ -26,11 +26,9 @@
     python_requires=">=3.7",
     install_requires=[
         "flatten-dict==0.4.2",
-        "jupyter==1.0.0",
         "matplotlib>=3.5.3",
```
kujon, line 29: Why is `matplotlib` required?

valeriiashmyhlo: Tests were failing, fixed it. `matplotlib` is optional now.

.circleci/config.yml
```diff
@@ -131,6 +131,8 @@ jobs:
             pip install -U pip
             pip install -e .
```
joloppo, line 132: This works, suggestion:

Use `pip install -e .[binder]` and add a "binder" section to the extras in `setup.py` much like the current `tests` section. Example in line 18 of this file where tests requirements are installed. you can remove all 3 (L 133, 134, 145) pip installs here after moving the requirements to the extra section.

setup.py
```diff
@@ -26,11 +26,8 @@
     python_requires=">=3.7",
     install_requires=[
         "flatten-dict==0.4.2",
-        "jupyter==1.0.0",
-        "matplotlib>=3.5.3",
         "pandas>=0.25.2",
         "requests==2.31.0",
-        "statsmodels==0.13.2",
         "tqdm==4.64.1",
         "pydantic==1.10.2",
```
joloppo, line 32: Not sure where we are using pydantic but binding it to a specific version is a bad idea.

setup.py
```diff
@@ -26,11 +26,8 @@
     python_requires=">=3.7",
     install_requires=[
         "flatten-dict==0.4.2",
-        "jupyter==1.0.0",
-        "matplotlib>=3.5.3",
         "pandas>=0.25.2",
         "requests==2.31.0",
-        "statsmodels==0.13.2",
         "tqdm==4.64.1",
```
joloppo, line 31: Not sure how we use tqdm but this could maybe be an extra as well.

valeriiashmyhlo (approved):

PR #431, opened 2023-04-20
feat: filter_time using now() as value

jakubbuskiewicz:
#### CHANGELOG

- changed default value for `filter_time_max` and `filter_time_min` for cargo movements to be consistent with API values (which is `now()`)

#### TESTS

- QA'd locally

#### COMMENTS

- any other business


vortexasdk/endpoints/cargo_movements.py
```diff
@@ -36,8 +36,8 @@ def __init__(self):
     def search(
         self,
         filter_activity: str = None,
-        filter_time_min: datetime = datetime(2019, 10, 1, 0),
-        filter_time_max: datetime = datetime(2019, 10, 1, 1),
```
ministrudels, line 40: Can we remove this entirely? The default is already set at the API level

jakubbuskiewicz: these fields are marked as required in our docs so I would keep it to be consistent

ministrudels (approved):

PR #439, opened 2023-08-25
fix: fixes for issues causing failing CICD pipeline

samhuk:
#### RELATED TICKETS

- https://vortexa.atlassian.net/browse/RND-3457

See ticket for detail.

### DIFF
--- a/tests/endpoints/test_onshore_inventories_timeseries.py
+++ b/tests/endpoints/test_onshore_inventories_timeseries.py
@@ -22,8 +22,6 @@ class TestOnshoreInventoriesTimeseries(TestCaseUsingRealAPI):
             .to_df()
         )
 
-        # This used to be 3, but the result from API has changed to contain 4.
-        # TODO: Investigate why. This is to fix pipeline issues.
         assert len(df) == 4
 
     def test_timeseries_breakdowns_by_country_should_contain_over_100_unique_splits(
--- a/tests/endpoints/test_onshore_inventories_timeseries.py
+++ b/tests/endpoints/test_onshore_inventories_timeseries.py
@@ -22,7 +22,9 @@ class TestOnshoreInventoriesTimeseries(TestCaseUsingRealAPI):
             .to_df()
         )
 
-        assert len(df) == 3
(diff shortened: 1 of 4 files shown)


tests/endpoints/test_voyages_congestion_breakdown.py
```diff
@@ -42,11 +42,12 @@ def test_search_returns_larger_size(self):
         start = datetime(2019, 1, 1)
         end = datetime(2019, 11, 10)
 
+        # Limit breakdown to 25 to prevent too many ES buckets down-the-line
```
migueluzcategui, line 45: I think that the problem is that we are using breakdown_port [here](https://github.com/VorTECHsa/python-sdk/blob/7553d7622df36929af3654b50a1b2ee216712604/vortexasdk/endpoints/voyages_congestion_breakdown.py#L30) and it should be one of these instead `vessel_count,utilisation,cargo_quantity,avg_wait_time,dwt,cubic_capacity,tonne_miles,avg_distance,avg_speed`

migueluzcategui (changes requested):

tests/endpoints/test_voyages_congestion_breakdown.py
ministrudels, line 45: Please remove this line

tests/endpoints/test_onshore_inventories_timeseries.py
```diff
@@ -22,7 +22,9 @@ def test_timeseries_returns_3_weeks_of_data(self):
             .to_df()
         )
 
-        assert len(df) == 3
+        # This used to be 3, but the result from API has changed to contain 4.
+        # TODO: Investigate why. This is to fix pipeline issues.
```
ministrudels, line 26: I think your fix is correct and we should remove this comment. With the changes to OnshoreInventories weeks, this used to return Thursday to Thursday weeks 5-11, 12-18, and 19-23. Now this would return Monday - Sunday weeks 2-8, 9-15, 16-22, 23-23

ministrudels (changes requested):

ministrudels (approved):

PR #465, opened 2024-01-22
Type hint Result().to_df() 

ministrudels:
#### CHANGELOG

This PR narrows interfaces and adds type annotations on commonly used utils 

- Removed `default_columns` parameter on the `create_dataframe` util for a narrower interface
- Type for `convert_to_flat_dict` 
- Type hint when calling `.to_df()` now suggests the `all` literal. This is not new functionality but was only documented in our sdk docs


migueluzcategui (approved):
LGTM

PR #493, opened 2024-04-19
Upgrades and Type Fixes

SeanBarry:
This is a colossal PR that changes _a lot_.

I wanted to avoid such a large PR but struggled to as there was a web of issues created by deprecated functionality or packages.

The main changes are:
- Upgrade package build to Python 3.11
- Upgrade minimum required Python version to 3.11. It was previously 3.7
- Increment major version (v1.0.0)
- Upgrading old packages
- Replacing deprecated functionality
- Fixing many type issues
- Fixing many lint/formatting issues


docs/examples/academy/15. Filtering and Modifying DataFrames.ipynb
```diff
@@ -1355,273 +271,25 @@
...
-     },
-     "execution_count": 12,
-     "metadata": {},
-     "output_type": "execute_result"
-    }
-   ],
+   "outputs": [],
    "source": [
-    "OPEC_df.groupby('Region').mean()"
+    "OPEC_df.groupby(\"Region\").mean(numeric_only=True)"
```
SeanBarry, line 292: This is the only meaningful change in this file - it fixes an issue where the `.mean` method was not working after a Pandas upgrade.

docs/examples/academy/28. Aggregating Cargo Movements Data.ipynb
```diff
@@ -314,23 +266,23 @@
...
      },
-     "execution_count": 9,
+     "execution_count": 11,
      "metadata": {},
      "output_type": "execute_result"
     }
    ],
    "source": [
-    "weekly_quantity = cms.groupby('loading_week').sum()\n",
+    "weekly_quantity = cms.groupby('loading_week').sum(numeric_only=True)\n",
```
SeanBarry, line 285: This is the only meaningful change in this file - it fixes an issue where the `.sum` method was not working after a Pandas upgrade.

docs/examples/academy/28. Aggregating Cargo Movements Data.ipynb
```diff
@@ -382,152 +327,173 @@
...
      },
-     "execution_count": 10,
+     "execution_count": 13,
      "metadata": {},
      "output_type": "execute_result"
     }
    ],
    "source": [
-    "quantity_by_category = cms.groupby(by = ['loading_week','product_category']).sum().reset_index()\n",
+    "quantity_by_category = cms.groupby(by = ['loading_week','product_category']).sum(numeric_only=True).reset_index()\n",
```
SeanBarry, line 496: This is a meaningful change in this file - it fixes an issue where the `.sum` method was not working after a Pandas upgrade.

docs/examples/academy/28. Aggregating Cargo Movements Data.ipynb
```diff
@@ -816,32 +768,18 @@
...
-     "iopub.status.idle": "2021-07-13T11:16:25.095115Z",
-     "shell.execute_reply": "2021-07-13T11:16:25.095523Z"
-    }
-   },
+   "execution_count": 17,
+   "metadata": {},
    "outputs": [],
    "source": [
-    "quantity_by_destination = cms.groupby('unloading_country').sum()[['quantity']]\n",
+    "quantity_by_destination = cms.groupby('unloading_country').sum(numeric_only=True)[['quantity']]\n",
```
SeanBarry, line 775: This is the only meaningful change in this file - it fixes an issue where the `.sum` method was not working after a Pandas upgrade.

docs/examples/academy/29. Working with Aggregated Cargo Movements Data.ipynb
```diff
@@ -95,41 +96,32 @@
...
-     "metadata": {
-      "needs_background": "light"
-     },
+     "metadata": {},
      "output_type": "display_data"
     }
    ],
    "source": [
-    "weekly_quantity = cms.groupby('loading_week').sum()\n",
+    "weekly_quantity = cms.groupby('loading_week').sum(numeric_only=True)\n",
```
SeanBarry, line 124: This is the only meaningful change in this file - it fixes an issue where the `.sum` method was not working after a Pandas upgrade.

docs/examples/academy/29. Working with Aggregated Cargo Movements Data.ipynb
```diff
@@ -147,41 +139,32 @@
...
-     "metadata": {
-      "needs_background": "light"
-     },
+     "metadata": {},
      "output_type": "display_data"
     }
    ],
    "source": [
-    "quantity_by_category = cms.groupby(by = ['loading_week','product_category']).sum().reset_index()\n",
+    "quantity_by_category = cms.groupby(by = ['loading_week','product_category']).sum(numeric_only=True).reset_index()\n",
```
SeanBarry, line 167: This is the only meaningful change in this file - it fixes an issue where the `.sum` method was not working after a Pandas upgrade.

docs/examples/academy/29. Working with Aggregated Cargo Movements Data.ipynb
```diff
@@ -247,39 +221,32 @@
...
+       "<Figure size 600x600 with 1 Axes>"
       ]
      },
      "metadata": {},
      "output_type": "display_data"
     }
    ],
    "source": [
-    "quantity_by_destination = cms.groupby('unloading_country').sum()[['quantity']]\n",
+    "quantity_by_destination = cms.groupby('unloading_country').sum(numeric_only=True)[['quantity']]\n",
```
SeanBarry, line 249: This is the only meaningful change in this file - it fixes an issue where the `.sum` method was not working after a Pandas upgrade.

docs/examples/academy/33. Analysing Imports Exports Data.ipynb
```diff
@@ -229,64 +228,62 @@
...
      },
-     "execution_count": 6,
+     "execution_count": 7,
      "metadata": {},
      "output_type": "execute_result"
     }
    ],
    "source": [
-    "imports_df.sum().sort_values(ascending=False).head(10)"
+    "imports_df.sum(numeric_only=True).sort_values(ascending=False).head(10)"
```
SeanBarry, line 256: This is the only meaningful change in this file - it fixes an issue where the `.sum` method was not working after a Pandas upgrade.

docs/examples/academy/33. Analysing Imports Exports Data.ipynb
```diff
@@ -229,64 +228,62 @@
...
-     "metadata": {
-      "needs_background": "light"
-     },
+     "metadata": {},
      "output_type": "display_data"
     }
    ],
    "source": [
-    "top10 = imports_df.sum().sort_values(ascending=False).head(10)\n",
+    "top10 = imports_df.sum(numeric_only=True).sort_values(ascending=False).head(10)\n",
```
SeanBarry, line 286: This is the only meaningful change in this file - it fixes an issue where the `.sum` method was not working after a Pandas upgrade.

CONTRIBUTING.md
```diff
@@ -22,40 +22,40 @@ ideal report includes:
...
-$ cd python-sdk
+git clone git@github.com:vortechsa/python-sdk.git
+cd python-sdk
 ```
 
 create a new virtual environment
 
 ```bash
-$ python3.7 -m venv venv
+python -m venv venv
```
matteo-pallini, line 32: nit: probably we should mention that when creating the virtual env they should use 3.11?

CONTRIBUTING.md
```diff
@@ -88,7 +88,7 @@ To run the SDK on an Apple ARM chip, upgrade your Python version to 3.8 or highe
 The set up is almost identical as above except one command; to create a virtual environment for Python 3.8 you should run:
```
matteo-pallini, line 88: remove reference to `python 3.8`. Also, are the new changes going to work on ARM chip? I think M1/2s may be using ARM. Should we ask anyone using them to double check?

mkdocs.yml
```diff
@@ -57,7 +57,6 @@ pages:
       Geography: entities/geography.md
       Product: entities/product.md
       Vessel: entities/vessel.md
-      Vessel Movement: entities/vessel_movement.md
```
matteo-pallini, line 60: nit: should we also release a minor version of the previous semver (using python 3.7) with this removed? so that whoever doesn't want to move to a higher pythong version still has available a fully working version of the SDK

matteo-pallini: or at least mark it as deprecated?

SeanBarry: This just removes a link to a non existent file in the docs. The link doesn't show because the file is non existent, so this change has no material effect on existing users.

setup.py
```diff
@@ -19,45 +19,45 @@
...
+        "Programming Language :: Python :: 3.11.0",
         "License :: OSI Approved :: Apache Software License",
         "Operating System :: OS Independent",
     ],
-    python_requires=">=3.7",
+    python_requires=">=3.11",
     install_requires=[
         "flatten-dict==0.4.2",
-        "pandas>=0.25.2",
+        "pandas==2.2.0",
```
matteo-pallini, line 29: This feels like a big jump. Should we also release an intermediate SDK version requiring pandas 1.*.* and possibly a python version in between? my concern is that we may end up forcing the SDK consumers either to live with a very old version of things either move everything up to something very new. 

Having said that this is only relevant if customers use the SDK in production code. If it's only ever used in notebooks then it's probably okay to bump everything up

tests/api/test_canal_transit_entity.py
```diff
@@ -1,31 +1,40 @@
...
-            serialized = json.load(f)
-            deserialized = CanalTransitRecord.parse_obj(serialized)
-
-            expected = CanalTransitRecord(
-                id="0004e233f16e15c0fcb3189332807e21306a3b3b06f0a778e3a3d9f276f233bd",
-                canal="canal1",
-                vessel_class=[
+    def test_serialize(self) -> None:
+        CanalTransitRecord.model_validate(
+            {
```
matteo-pallini, line 9: nit: to keep the test closer to how it was should we pass a string as an argument and used `model_validate_json` instead?

tests/api/test_cargo_event_entity.py
```diff
@@ -1,33 +1,26 @@
...
-                location=[
-                    GeographyEntity(
-                        id="2aaad41b89dfad19e5668918018ae02695d7710bcbe5f2dc689234e8da492de3",
-                        layer="country",
-                        label="United Kingdom",
-                        source="model",
-                        probability=1,
-                    )
+    def test_serialize(self) -> None:
+        CargoPortUnloadEvent.model_validate(
```
matteo-pallini, line 8: same here https://github.com/VorTECHsa/python-sdk/pull/493/files#r1575914768

vortexasdk/__init__.py
```diff
@@ -46,3 +46,46 @@
 
 # noinspection PyUnresolvedReferences
 from vortexasdk.check_setup import run_all_checks
+
+# Explicitly define the exported symbols
+__all__ = [
```
matteo-pallini, line 51: Following this change should we also update other documentation anywhere else to list the objects available?

vortexasdk/api/cargo_movement.py
```diff
@@ -18,7 +18,7 @@ class RawLocations(BaseModel):
 class CargoPortLoadEvent(BaseModel):
     vessel_id: Optional[ID] = None
     start_timestamp: Optional[ISODate] = None
-    event_type: Optional[Literal["cargo_port_load_event"]] = None
+    event_type: Literal["cargo_port_load_event"]
```
matteo-pallini, line 21: This is probably something we may want to add in a release changelog. Let's not forget about this

vortexasdk/api/geography.py
```diff
@@ -25,10 +25,10 @@ class Geography(Node):
 
     id: ID
     name: Optional[str] = None
-    layer: Optional[List[str]] = None
+    layer: List[str] = []
     exclusion_rule: Optional[List[IDNameLayer]] = None
     hierarchy: Optional[List[IDLayer]] = None
-    pos: Optional[List[str]] = None
+    pos: Optional[List[str | float]] = None
```
matteo-pallini, line 31: how come this can be both a string and float? it's a bit of an odd signature, no?

vortexasdk/api/search_result.py
```diff
@@ -15,22 +15,24 @@ def to_list(self) -> List:
         """Represent *records* as a list."""
         return self.records
 
-    def to_df(self, columns: Union[Literal["all"], List[str]]) -> pd.DataFrame:
+    def to_df(
+        self, columns: Literal["all"] | List[str] | None
```
matteo-pallini, line 19: I assume `None` and `Literal["all"]` are going to behave in the same way? could we remove `Literal["all"]` maybe? or should we document this in the docstring?

vortexasdk/api/voyages.py
```diff
@@ -122,7 +122,7 @@ class VoyageCargoEvent(BaseModel):
     odometer_start: Optional[int] = None
     odometer_end: Optional[int] = None
     cargo_destination_id: Optional[str] = None
-    tonne_miles: Optional[int] = None
+    tonne_miles: Optional[int | float] = None
```
matteo-pallini, line 125: how come can this have both types now?

vortexasdk/client.py
```diff
@@ -36,7 +36,7 @@ class VortexaClient:
     _N_THREADS = 6
     _MAX_ALLOWED_TOTAL = int(1e6)
 
-    def __init__(self, **kwargs):
+    def __init__(self: "VortexaClient", **kwargs: Dict) -> None:
```
matteo-pallini, line 39: nit: should we have `Any` rather than `Dict` here? or are we expecting individual kwargs to be only dict ?https://peps.python.org/pep-0484/#arbitrary-argument-lists-and-default-argument-values

vortexasdk/utils.py
```diff
@@ -225,3 +227,17 @@ def sts_param_value(param):
...
         return {"exclude": False, "x_filter": False}
+
+
+class PAGINATION_STRATEGIES(Enum):
+    OFFSET = "OFFSET"
+    SEARCH_AFTER = "SEARCH_AFTER"
+
+
+def showDeprecatedGeoExclusionRulesWarning(
+    disable_geographic_exclusion_rules: Optional[bool], logger: Logger
```
matteo-pallini, line 238: nit: should we default it to `False` and type it as `bool` rather than having it as Optional?

matteo-pallini (commented):
Left a bunch of comments, most of them (if not all of them) are nit, so feel free to ignore them. The only significant one is probably this one https://github.com/VorTECHsa/python-sdk/pull/493#discussion_r1575907490

setup.py
SeanBarry: There are quite a few big jumps in this PR. I personally don't think it's a problem. Generally the API of packages doesn't change _that_ much over time, and I only needed to make a handful of changes in this codebase to fix any Pandas issues caused by the upgrade.

I personally think we don't need to introduce intermediary versions. Our clients can continue to use an older version of the SDK if they wish. Happy to open this Q up to a wider audience though.

tests/api/test_cargo_event_entity.py
SeanBarry: I think the test is easier to understand and work on when the actual test data is defined in the same file. I don't think this change has any material effect other than readability?

vortexasdk/api/cargo_movement.py
SeanBarry: These types are deceiving.
In the Python SDK almost every attribute in every model is `Optional`. This does not reflect reality, instead it was a decision made to ensure that in the unlikely event a breaking change was made in the API, it would not break pipelines created by SDK consumers. 
In this particular instance, this field is being used as a discriminator ([Link](https://docs.pydantic.dev/latest/concepts/unions/#discriminated-unions)) and the new version of Pydantic didn't like that it was optional.
In reality, this key is not nullable, so there is no actual change here or risk to SDK users.

vortexasdk/api/search_result.py
SeanBarry: Good catch, I actually needed to explicitly handle ensuring that the behaviour was the same for `all` and `None`. I've made this change.

vortexasdk/api/voyages.py
SeanBarry: No, we are not. Fixed!

setup.py
matteo-pallini: I guess that if it is a big problem they will raise it anyway :-)

thomasfrederikhoeck: I would really recommend not going back to a situation where pydantic force a specific version of `pydantic` but also `pandas` (previouslt it was fixed but was relaxed when we reached out). In general as a library it is not recommend to create hard-pinned versions since it can it imposible to fit into an enviorment with packages that have other requirements. Only applications should hard-pin. See https://stackoverflow.com/a/44938662 for an example.

SeanBarry: Hey @thomasfrederikhoeck, thanks for reaching out. This PR was an attempt to fix some outdated packages and bring the developer experience for the SDK up to scratch. Due to the number of changes and the potential risk to users of the SDK we decided not to merge it yet. I've noted your points about pinning specific versions. I agree this is not ideal. 
I'll be revisiting this PR soon to try to tidy it up and address the point you've raised. Cheers

SeanBarry:
Superseded by https://github.com/VorTECHsa/python-sdk/pull/526

PR #500, opened 2024-05-30
fix: check if key col exists

ministrudels:
#### RELATED TICKETS

- asana/app-issues...

#### RELATED PRs

- web/app-core/publish...

#### CHANGELOG

- describe code changes

#### TESTS

- describe your test strategy

#### COMMENTS

- any other business


joloppo (approved):
Ah, I thought I broke it. This was already a bug before. I had discussed this with jb previously. Technically the timeseries api should never return an empty response.

migueluzcategui (approved):

joloppo:
The result of a timeseries request should always be the length of the requested timeseries. So if we want to fix this client-side, we should generate some df with a bunch of 0's.

PR #539, opened 2025-01-13
fix: raise exception when response has an error

cojdev:
#### RELATED TICKETS

- https://vortexa.atlassian.net/browse/RND-11798

#### CHANGELOG

- when an error is returned from a `search_base` request, an exception is now raised with that error.

#### TESTS

- tested locally

#### COMMENTS

- any other business


vortexasdk/client.py
```diff
@@ -82,8 +82,14 @@ def search_base(
         probe_response = _send_post_request(
             url, payload, size=1, headers=headers
         )
+
+        if "error" in probe_response:
+            raise Exception(probe_response)
+
         total = self._calculate_total(probe_response)
 
+        print(probe_response)
```
migueluzcategui, line 91: Are you sure about this `print` here? For large responses, it could add significant performance overhead and make the logs unreadable. What if you log just the total number of records instead, response time and so? I think that info is more usable in most of the cases

cojdev: ah, that's actually leftover from debugging, I'll remove it

migueluzcategui (approved):

PR #546, opened 2025-05-07
Support parsing different ISO compliant strings

dhruvsegat-vortexa:
#### RELATED TICKETS

- https://vortexa.atlassian.net/browse/RND-14151

#### CHANGELOG

- Adds support for parsing different ISO format datetime strings in a series

#### TESTS

- Added tests


tests/test_result_conversions.py
```diff
@@ -0,0 +1,69 @@
...
+from vortexasdk.result_conversions import format_datatypes
+
+
+def test_format_datatypes():
+    # Create a sample DataFrame with different timestamp formats
+    data = {
+        "date_col_dont_convert": ["2023-10-01", "2023-10-02", "2023-07-01"],
+        "date_col_timestamp": ["2023-10-01", "2023-10-02", "2023-07-01"],
+        "mixed_iso_format_timestamps": [
+            "2023-10-03 12:00:00",
```
maxhipperson, line 14: These aren’t ISO format date times. If they’re missing the T. What are these test cases supposed to be?

maxhipperson (changes requested):

tests/test_result_conversions.py
maxhipperson: Handled in slack. They’re formats other than 8601. I assumed we stuck with the latest, but that’s not necessarily true.

tests/test_result_conversions.py
```diff
@@ -0,0 +1,69 @@
+from datetime import datetime
+
+import pandas as pd
+
+from vortexasdk.result_conversions import format_datatypes
+
+
+def test_format_datatypes():
```
maxhipperson, line 8: Can we split these test cases into different tests or parameteristations? Then it’ll be easier to understand what fails when. It’s much harder when the test cases are parametrised within a data frame .

dhruvsegat-vortexa: Done

jakubbuskiewicz (approved):

vortexasdk/result_conversions.py
```diff
@@ -26,7 +26,14 @@ def format_datatypes(df: pd.DataFrame) -> pd.DataFrame:
     timestamp_cols = [col for col in df.columns if "timestamp" in col]
 
     for col in timestamp_cols:
-        df[col] = pd.to_datetime(df[col])
+        if df[col].dtype != "object":
```
matteo-pallini, line 29: any chance we might want to do the same here as well? https://github.com/VorTECHsa/python-sdk/blob/9718939eefa1a8d12415367985b427931234b146/vortexasdk/endpoints/timeseries_result.py#L102

vortexasdk/result_conversions.py
```diff
@@ -26,7 +26,14 @@ def format_datatypes(df: pd.DataFrame) -> pd.DataFrame:
...
+            df[col] = pd.to_datetime(df[col])
+        else:
+            try:
+                df[col] = pd.to_datetime(df[col], format="ISO8601")
+            except (ValueError, pd.errors.ParserError):
+                logger.debug("Failed to parse ISO8601 format, trying default")
```
matteo-pallini, line 35: nit: I would probably also log the column name here

vortexasdk/result_conversions.py
```diff
@@ -26,7 +26,14 @@ def format_datatypes(df: pd.DataFrame) -> pd.DataFrame:
     timestamp_cols = [col for col in df.columns if "timestamp" in col]
 
     for col in timestamp_cols:
-        df[col] = pd.to_datetime(df[col])
+        if df[col].dtype != "object":
+            df[col] = pd.to_datetime(df[col])
+        else:
+            try:
+                df[col] = pd.to_datetime(df[col], format="ISO8601")
+            except (ValueError, pd.errors.ParserError):
```
matteo-pallini, line 34: As far as I can tell the tests are not going through this branch, something that we probably want?
Screenshot 2025-05-08 at 12 13 49

matteo-pallini (commented):
Looks good, and I am fairly confident it would work. Although, if this is correct it might be something worth addressing given the relevance of this library?https://github.com/VorTECHsa/python-sdk/pull/546/files#r2079513709

vortexasdk/result_conversions.py
dhruvsegat-vortexa: Done, added a non ISO format (`YYYY/mm/dd`) test case

vortexasdk/result_conversions.py
dhruvsegat-vortexa: I'm a bit conflicted on this one. Although I'm confident this shouldn't cause any regression, I'd prefer to put this out in the wild, make sure its working and then change other usages of `pd.to_datetime`. Plus no bugs have been reported around this one.

matteo-pallini (approved):
LGTM

PR #564, opened 2026-05-20
Add AFP methods to SDK

Jack-Burge55:
#### RELATED TICKETS

- https://vortexa.atlassian.net/browse/RND-21261

! Do not post related PRs here, unless they are open source !

#### CHANGELOG

- Added SDK methods for the following AFP endpoints:
- https://api.vortexa.com/v5/anywhere-freight-pricing/latest-update-timestamp
- https://api.vortexa.com/v5/anywhere-freight-pricing/price-timeseries
- https://api.vortexa.com/v5/anywhere-freight-pricing/price/details (GET + POST)
- https://api.vortexa.com/v5/anywhere-freight-pricing/top-ports/destination
- https://api.vortexa.com/v5/anywhere-freight-pricing/top-ports/origin
- https://api.vortexa.com/v5/anywhere-freight-pricing/vessel-classes-details

#### TESTS

- Added tests for each added method, and spun up a notebook to test each one

#### COMMENTS

- None


vortexasdk/endpoints/anywhere_freight_pricing_result.py
```diff
@@ -0,0 +1,30 @@
+from typing import Any, Dict, List
+
+import pandas as pd
+from pydantic import BaseModel, Field
+
+
+class AnywhereFreightPricingResult(BaseModel):
```
samhuk: This class re-implements `records`, `reference`, `to_list`, and `to_df` from scratch instead of subclassing `vortexasdk.api.search_result.Result` like every other `*Result` in this repo (e.g. `FreightPricingResult`, `RefineriesResult`, `TimeSeriesResult`). As written, instances won't have `__len__`, `__iter__`, `__getitem__`, or `__str__`, and `to_df` drops the `columns` argument that callers of other endpoints can rely on. Could we extend `Result` here for consistency? If the base `to_df` doesn't fit the json-normalize approach, override it but keep the same signature.

vortexasdk/endpoints/anywhere_freight_pricing_get_price_details.py
```diff
@@ -0,0 +1,141 @@
...
+    def __init__(self) -> None:
+        self._resource = ANYWHERE_FREIGHT_PRICING_PRICE_DETAILS
+
+    def search(
+        self,
+        time_min: datetime,
+        time_max: datetime,
+        origin_port: str,
+        destination_port: str,
+        vessel_class: str,
```
samhuk: Existing endpoints in this repo (e.g. `Refineries.search(status: Optional[Literal["Active", "Shut", "Unknown", "Upcoming"]])`) constrain enumerated params with `Literal` so callers get IDE/mypy validation. The AFP endpoints document the allowed values for `vessel_class`, `product`, `unit`, `frequency`, and `avoid_zone` items but type them as plain `str`/`List[str]`. Worth lifting these into module-level `Literal` aliases (e.g. `VesselClass = Literal["oil_coastal", "oil_specialised", ...]`) and reusing them across the AFP endpoints. The `routes` dicts in the POST endpoints would benefit from a `TypedDict` for the same reason.

vortexasdk/endpoints/anywhere_freight_pricing_post_price_details.py
```diff
@@ -0,0 +1,117 @@
...
+    AnywhereFreightPricingResult,
+)
+from vortexasdk.logger import get_logger
+from vortexasdk.operations import Search
+from vortexasdk.utils import to_date_string
+
+logger = get_logger(__name__)
+
+
+class AnywhereFreightPricingPostPriceDetails(Search):
```
samhuk, line 26: This class and `AnywhereFreightPricingGetPriceDetails` both target `ANYWHERE_FREIGHT_PRICING_PRICE_DETAILS` (`/v5/anywhere-freight-pricing/price/details`), one via GET and one via POST. The repo convention (e.g. `Refineries(Reference, Search)`) is one class per resource exposing multiple methods. Could these be merged into a single `AnywhereFreightPricingPriceDetails` with `.search(...)` (single route, GET) and `.search_multi(routes=[...])` (multi-route, POST)? Two top-level imports for the same logical operation is awkward at the public-API level.

vortexasdk/endpoints/anywhere_freight_pricing_get_price_details.py
```diff
@@ -0,0 +1,141 @@
...
+
+        if suggested_tonnage is not None:
+            params["suggested_tonnage"] = suggested_tonnage
+
+        client = default_client()
+        url = client._create_url(self._resource)
+
+        # Use doseq=True to handle list parameters (avoid_zone)
+        query_string = urlencode(params, doseq=True)
+        url = f"{url}&{query_string}"
```
samhuk: Unlike the POST/timeseries variants, this class doesn't extend `Search` and instead hand-rolls URL construction:
```python
url = client._create_url(self._resource)
query_string = urlencode(params, doseq=True)
url = f"{url}&{query_string}"
```
That works (because `_create_url` already adds `?_sdk=...&apikey=...`), but it duplicates and slightly bypasses `_create_url_with_params`. Consider extending `_create_url_with_params` to accept `doseq=True` for list params, then call it here — keeps the single source of truth for URL composition and apikey handling.

vortexasdk/endpoints/anywhere_freight_pricing_get_price_details.py
```diff
@@ -0,0 +1,141 @@
...
+            "time_min": to_date_string(time_min),
+            "time_max": to_date_string(time_max),
+            "origin_port": origin_port,
+            "destination_port": destination_port,
+            "vessel_class": vessel_class,
+            "product": product,
+            "unit": unit,
+        }
+
+        if avoid_zone:
```
samhuk: Mixed conventions for optional params across the new files: this endpoint uses `if avoid_zone:` and `if suggested_tonnage is not None:`, while `top_ports_*` uses `if avoid_zone is not None:`. The first form drops empty lists (intentional or not?), the second keeps them. Pick one (`is not None` is the safer default) and apply uniformly across all AFP endpoints.

vortexasdk/utils.py
```diff
@@ -93,6 +93,11 @@ def filter_empty_values(data: Dict) -> Dict:
     }
 
 
+def to_date_string(dt: datetime) -> str:
```
samhuk: `to_date_string` overlaps with the existing `to_ISODate` in `vortexasdk/api/shared_types.py`, just with a different format. If the AFP API truly requires `YYYY-MM-DD` rather than ISO-8601 with time, please add a comment naming the spec, and consider co-locating it with `to_ISODate` in `shared_types.py` instead of `utils.py` so the two date helpers are discoverable together.

mkdocs.yml
```diff
@@ -24,6 +24,16 @@ pages:
       Products: endpoints/products.md
       Vessels: endpoints/vessels.md
       Fixtures: endpoints/fixtures.md
+      Anywhere Freight Pricing:
+        Latest Update: endpoints/anywhere_freight_pricing_latest_update_timestamp.md
+        Time-series: endpoints/anywhere_freight_pricing_price_timeseries.md
+        Price:
+          Get Prices: endpoints/anywhere_freight_pricing_get_price_details.md
+          Post Prices: endpoints/anywhere_freight_pricing_post_price_details.md
+        Top Ports:
```
samhuk: `pydocmd.yml` generates `endpoints/anywhere_freight_pricing_result.md` but it's not listed under `pages:` here. Either drop the generation entry (the result class is internal) or add a Result page to the AFP menu so the doc isn't orphaned. Also, this line has trailing whitespace after `Top Ports:`.

vortexasdk/endpoints/anywhere_freight_pricing_latest_update_timestamp.py
```diff
@@ -0,0 +1,58 @@
+from typing import Dict
```
samhuk, line 6: Every other endpoint module in `vortexasdk/endpoints/` opens with a Binder badge docstring linking to a try-me-out notebook. The seven new AFP modules omit it. If notebooks aren't ready yet, fine to defer — flagging for parity.

vortexasdk/endpoints/anywhere_freight_pricing_post_price_details.py
Jack-Burge55: I'm not sure if we should combine these two. This may be the convention in this repo (which is a little left behind) but our current documentation for endpoints has these as separate docs, one for each endpoint. Don't you think we should stick with that to be consistent?

samhuk (approved):

